PR #1, opened 2022-07-15
Add initial gha-tools files

sevagh:
For an example of these tools being used to upload and download wheels, see the [WIP RMM cibuildwheel workflow](https://github.com/rapidsai/rmm/pull/1070)


tools/rapids-upload-to-anaconda
```diff
@@ -0,0 +1,54 @@
...
+  *)
+    echo 'Only branch builds and nightly builds are uploaded to Anaconda.org'
+    exit 1
+    ;;
+esac
+
+CONDA_USERNAME="rapidsai-nightly"
+CONDA_TOKEN="${NIGHTLY_CONDA_TOKEN}"
+if [[ "${GIT_BRANCH}" == "main" ]]; then
+  CONDA_USERNAME="rapidsai"
```
raydouglass, line 41: You should be able to override this. Some of our packages are uploaded to the `nvidia` channel from the main branch.

raydouglass (changes requested):

tools/rapids-upload-to-anaconda
ajschmidt8: I wrote this script initially. In hindsight, do we need to keep around the `CONDA_USERNAME` variable at all? It's only used in the `-u` flag for the `anaconda upload` command. And from the docs below, it says the following about the `-u` flag usage:

```
-u USER, --user USER  User account or Organization, defaults to the current
                        user
```

- https://docs.anaconda.com/anacondaorg/commandreference/#upload

If the upload command defaults to the current user, why do we need to override it at all with the `-u` flag?

tools/rapids-bash-functions
```diff
@@ -0,0 +1,3 @@
+function echo_stderr {
+    echo "    [$(basename $0)] $*" >&2
```
ajschmidt8, line 2: can you add a comment for why `basename $0` is needed? seems ambiguous to me. this file is ultimately sourced by another file, right? 

maybe add a comment header to this file entirely to state it's intended use

tools/rapids-s3-path
```diff
@@ -0,0 +1,136 @@
...
+    conda_python)
+      append_arch=1
+      append_pyver=1
+      ;;
+    wheel_python)
+      append_arch=1
+      append_pyver=1
+      ;;
+    *)
+      echo "Using input '${pkg_type}' as a custom file" >&2
```
ajschmidt8, line 103:
```suggestion
      echo_stderr "Using input '${pkg_type}' as a custom file"
```

tools/rapids-s3-path
```diff
@@ -0,0 +1,136 @@
...
+  # for cpp or python package types, its a tarball, append .tar.gz and prepend project name
+  if (( $is_tarball == 1 )); then
+    pkg_type="${repo_name}_${pkg_type}.tar.gz"
+  fi
+
+  # misc package types just have the package name pass through unchanged
+  echo -n "${pkg_type}"
+}
+
+PKG_TYPE="$(validate_package_type $1)"
```
ajschmidt8, line 132: this is trivial and i'm fine if you leave it as is, but it seems the `PKG_TYPE` is actually a filename, right?

does it make sense to rename this variable and function to reflect something about the filename?

tools/rapids-s3-path
```diff
@@ -0,0 +1,136 @@
...
+  fi
+
+  s3_directory_id=""
+  s3_prefix=""
+
+  case "${build_type}" in
+    pull-request)
+      # For PRs, $GITHUB_REF_NAME is like:
+      # 1058/merge
+      s3_directory_id="${GITHUB_REF_NAME%%/*}"
```
ajschmidt8, line 52: All pull requests will be copied to a branch in the source repository that has the format `pull-request/<PR_NUMBER>`.

This logic will need to be updated to reflect that. This aligns with our discussions from the CircleCI article below about _Marking code as trusted by pushing to upstream_.

https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/

In essence, technically PR builds will still be branch builds.

tools/rapids-upload-misc-to-s3
```diff
@@ -0,0 +1,30 @@
...
+  exit 1
+fi
+
+FILE_TO_UPLOAD="$1"
+ARTIFACT_NAME="$2"
+
+if [[ -d "${FILE_TO_UPLOAD}" ]]; then
+  echo_stderr "Misc/one-off uploads support files only, not directories"
+  exit 1
+fi
```
ajschmidt8, line 21: I think we should add an additional check somewhere around here to ensure that the file exists too. Right now we're just checking that it's not a directory.

tools/rapids-upload-misc-to-s3
```diff
@@ -0,0 +1,30 @@
...
+
+# Generate destination S3 path
+S3_DEST_PATH=$(rapids-s3-path "${ARTIFACT_NAME}")
+
+echo_stderr "Uploading ${FILE_TO_UPLOAD} to ${S3_DEST_PATH}"
+
+aws s3 cp --only-show-errors "${FILE_TO_UPLOAD}" "${S3_DEST_PATH}"
+
+echo_stderr "Browse uploaded file: ${S3_DEST_PATH/s3:\/\/rapids-downloads\//https:\/\/downloads.rapids.ai\/}"
```
ajschmidt8, line 30: seems like this `s3://...` to `https://...` conversion happens in a few spots. Can you consolidate this logic? Maybe in `rapids-bash-functions`?

You could actually probably just make a function like `print-browser-url` or something similar and pass it a filename, where it can handle the conversion and printing to `stderr` altogether.

tools/rapids-upload-to-anaconda
```diff
@@ -0,0 +1,54 @@
...
+  CONDA_TOKEN="${STABLE_CONDA_TOKEN}"
+fi
+
+echo "Uploading packages to Anaconda.org: ${PKGS_TO_UPLOAD}" 1>&2 # echo to stderr
```
ajschmidt8, line 45:
```suggestion
echo_stderr "Uploading packages to Anaconda.org: ${PKGS_TO_UPLOAD}"
```

tools/rapids-upload-to-s3
```diff
@@ -0,0 +1,24 @@
...
+  exit 1
+fi
+
+PATH_TO_TAR_UP="$1"
+S3_DEST_PATH="$2"
+
+echo_stderr "Compressing and uploading ${PATH_TO_TAR_UP} to ${S3_DEST_PATH}"
+
+# Tar.gz to stdout and pipe that to the s3 command, no need for a local file
+tar czfv - -C "${PATH_TO_TAR_UP}" . | aws s3 cp --only-show-errors - "${S3_DEST_PATH}"
```
ajschmidt8, line 21: If we want to leverage this file for `rapids-upload-misc-to-s3`, we could do something like the following in this script:

- Check `$1` to see if it's a file or a directory
- If it's a file, just copy it as-is to `$2`
- Else, tar it up and then pipe the output to `aws s3 cp ...` like you're already doing.

Up to you. If we adopt this method though, we could consolidate the `Browse parent directory of uploaded file` bits to a single file.

ajschmidt8 (commented):
Awesome, looks good. Left a few thoughts.

tools/rapids-s3-path
sevagh: I changed this function from `validate_package_type` to `format_package_name`, and manipulate a variable called `pkg_name=${pkg_type}` instead to create `PKG_NAME` in the end.

tools/rapids-upload-to-anaconda
sevagh: Interactions between `-t TOKEN` and `upload -u USER`: https://github.com/Anaconda-Platform/anaconda-client/issues/456

sevagh: We decided that:
* We won't do any logic to set or decide CONDA_TOKEN in the script
* The caller must set `CONDA_TOKEN=${whatever_i_want}`. The vision would be that we could have RAPIDSAI organization secrets such as `NVIDIA_CONDA_TOKEN`, `RAPIDS_NIGHTLY_TOKEN`, etc., each one corresponding to the appropriate channel on Anaconda
* We won't pass any override of `-u ${CONDA_USER}`, and instead use the user from the specified auth token

tools/rapids-s3-path
sevagh: I changed the logic to the following:
* Add a comment describing the above trusted code situation and how we copy PRs to branches named pull-request in the case of BUILD_TYPE="pull-request"
* Extract the part of the branch name trailing after the `/`, e.g. `BUILD_TYPE="pull-request", GITHUB_REF_NAME="pull-request/1058"` -> `s3_directory_id="1058"`

.github/workflows/release.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Package Scripts
+        run: |
+          tar -czf ${{ env.RELEASE_FILE_NAME }} -C ./tools .
+      - name: Compute Next Tag Name
+        run: |
+          LAST_TAG=$(git describe --tags --abbrev=0)
```
ajschmidt8, line 27:
```suggestion
          LAST_TAG=$(git describe --tags --abbrev=0 || echo "0.0.0")
```

I know this was copied from the other repo. I think it will fail since no tags currently exist in the repository. We might want to add an `||` statement to account for this.

raydouglass (approved):

ajschmidt8 (approved):
Nice! LGTM. Just one comment about the GitHub Actions for these repo.

PR #3, opened 2022-09-07
Add rapids-check-env tool

Ethyling:
Add `rapids-check-env` tool to `gha-tools`

### DIFF
--- a/tools/rapids-check-env
+++ b/tools/rapids-check-env
@@ -6,7 +6,7 @@
 #
 
 rapids-logger "Get env"
-env
+env | sort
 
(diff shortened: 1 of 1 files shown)


tools/rapids-check-env
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#
+# rapids-check-env
+#
+# prints summary of environment
+#
+
+rapids-logger "Get env"
+env
```
ajschmidt8, line 9:
```suggestion
env | sort
```

I think this would be useful.

ajschmidt8 (approved):

PR #10, opened 2022-09-21
Add RAPIDS_ARCH to override host arch

sevagh:
In cibuildwheel's publish step, I found it wasteful to have to launch separate x86_64 and aarch64 runners to simply download and re-upload a built wheel.

With this change, a runner can force a specific architecture of a package name, so one runner can download both architectures.

Example:
```
RAPIDS_ARCH="x86_64" rapids-download-wheels-from-s3 ./dist
RAPIDS_ARCH="aarch64" rapids-download-wheels-from-s3 ./dist
twine upload ./dist/*.whl
```

### DIFF
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -56,9 +56,9 @@ if (( append_pyver == 1 )); then
 fi
 
 # for cpp and python package types, always append arch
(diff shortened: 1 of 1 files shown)


Ethyling:
I'm curious about the `RAPIDS_PY_WHEEL_ARCH` name. Do we agree this apply to any kind of "package"? Should we just name it `RAPIDS_ARCH`?

Ethyling (approved):

PR #20, opened 2022-10-06
Feat/pypi wheel scripts

sevagh:
This adds a new Python script, `rapids_twine.py`, for uploading packages to Twine with a custom build tag from the cibuildwheel workflow.

This lets us overwrite same-versioned wheels using a simple integer tag in the wheel file name through the `RAPIDS_PY_WHEEL_BUILD_TAG` env var.

### DIFF
--- a/.github/workflows/prs.yaml
+++ b/.github/workflows/prs.yaml
@@ -19,6 +19,7 @@ jobs:
         uses: ludeeus/action-shellcheck@master
         with:
           scandir: "./tools"
+          ignore_paths: "**.py"
       - name: Check Permissions
         run: |
           # Executable bit indices
--- a/tools/rapids-twine.py
+++ b/tools/rapids-twine.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+# A utility script that uploads wheels to pip indexes with twine to support build numbers on conflict
+# See:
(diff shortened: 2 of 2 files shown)


ajschmidt8 (approved):
left this comment in Slack, but I'll post it here too. otherwise, LGTM.


do you prefer having the .py extension on that script? I think you could drop it if you'd prefer so that invoking it looks like the other scripts. but if you prefer it that's fine too.

sevagh:
By leaving the py extension we can skip Python scripts from shellcheck: https://github.com/rapidsai/gha-tools/pull/20/files#diff-b92d47056fd4c023911c36953668d69f6b8b344806e837f7c699db6ac2e751e6

ajschmidt8:
> By leaving the py extension we can skip Python scripts from shellcheck: https://github.com/rapidsai/gha-tools/pull/20/files#diff-b92d47056fd4c023911c36953668d69f6b8b344806e837f7c699db6ac2e751e6

I see. I assumed that was why. I'm indifferent. FWIW, it looks like we can specify multiple files for shellcheck. But that requires us to maintain the list, which could be a pain. either way is fine to me :+1: 

https://github.com/ludeeus/action-shellcheck#ignore-paths-and-names

PR #23, opened 2022-10-28
Append CUDA version to conda packages

Ethyling:
We need to add CUDA version to conda package names so we can upload both CUDA 11 and 12 packages without conflict on the name. This will append `cuda11` or `cuda12`

### DIFF
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -15,13 +15,16 @@ fi
 
 pkg_type="$1"
 
+append_cuda=0
 append_pyver=0
 append_wheelname=0
 
 case "${pkg_type}" in
   conda_cpp)
+    append_cuda=1
     ;;
   conda_python)
+    append_cuda=1
     append_pyver=1
     ;;
   wheel_python)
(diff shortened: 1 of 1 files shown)


leofang:
May I ask why is it needed? Conda can have packages built against different CUDA versions happily coexist, e.g. [CuPy](https://github.com/conda-forge/cupy-feedstock) is now supported with CUDA 10.2 & 11.x. If this is only for human readability, we'd add CUDA version to the build string, like what [PyTorch](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/103634438261898bd3617336ad3594f831f31278/recipe/meta.yaml#L36) does.

ajschmidt8:
> May I ask why is it needed? Conda can have packages built against different CUDA versions happily coexist, e.g. [CuPy](https://github.com/conda-forge/cupy-feedstock) is now supported with CUDA 10.2 & 11.x. If this is only for human readability, we'd add CUDA version to the build string, like what [PyTorch](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/103634438261898bd3617336ad3594f831f31278/recipe/meta.yaml#L36) does.

@leofang, this change does not affect the naming of the conda packages that are published to Anaconda.org. It only changes of the names of the intermediate CI tarball artifacts that are used throughout our CI workflows. The CUDA 11 and 12 workflows will be distinct, so we need to ensure that the artifacts are labeled appropriately.

ajschmidt8 (changes requested):
On second thought, we'll need to also add the CUDA version to the Python conda artifacts

ajschmidt8 (approved):
whoops. nevermind. i see you did that already.

leofang:
Thanks, AJ 🙂

PR #24, opened 2022-11-04
Update `rapids-dependency-file-checker`

ajschmidt8:
This PR updates `rapids-dependency-file-checker` to ensure that it exits with a zero code if there are no existing generated dependency files that need to be deleted.

This became in issue in `rapids-cmake` since that repository uses a `dependencies.yaml` file, but does not write any files to the repository (e.g. there are no `conda/environments/*` files that are generated).

The error message can be seen in the GitHub Action log below for https://github.com/rapidsai/rapids-cmake/pull/283.

- https://github.com/rapidsai/rapids-cmake/actions/runs/3396159139/jobs/5646913087

### DIFF
--- a/tools/rapids-dependency-file-checker
+++ b/tools/rapids-dependency-file-checker
@@ -12,7 +12,7 @@ grep -rlP \
   --include="*.txt" \
   --include="*.yaml" \
   "${SEARCH_PATTERN}" . | \
-  xargs rm
+  xargs rm || true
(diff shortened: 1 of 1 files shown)


ajschmidt8:
I tested this locally and it worked as intended

sevagh (approved):

raydouglass (approved):

PR #28, opened 2022-11-15
Feat/pip wheel version script

sevagh:
This PR adds a new script, `rapids-pip-wheel-version`, that generates pip wheel version strings by using the git tag, stripping alphabet characters, and appending an epoch timestamp. Currently being used successfully for nightly wheels in the pip wheel project.

It also replaces the overly-complicated rapids-twine.py with a bash equivalent rapids-twine.


tools/rapids-twine
```diff
@@ -0,0 +1,38 @@
...
+# A utility script that wraps twine to support build tags (PEP427)
+#        
+# Positional Arguments:
+#   1) wheel dir
+set -exo pipefail
+export RAPIDS_SCRIPT_NAME="rapids-twine"
+
+rapids-require-env-var "TWINE_USERNAME"
+rapids-require-env-var "TWINE_PASSWORD"
+rapids-require-env-var "TWINE_REPOSITORY_URL"
```
ajschmidt8, line 11: I removed `rapids-require-env` in lieu of `set -u` in the PR below

- https://github.com/rapidsai/gha-tools/pull/27

ajschmidt8 (approved):
one comment, otherwise LGTM

tools/rapids-twine
ajschmidt8: based on the arguments in the docs below, you can probably get away with just adding the following `twine` flags with `set -u` to accomplish the same thing:

```sh
twine upload \
  --username "$TWINE_USERNAME" \
  --password "$TWINE_PASSWORD" \
  --repository-url "$TWINE_REPOSITORY_URL"
```


- https://twine.readthedocs.io/en/stable/#twine-upload

PR #35, opened 2023-01-23
Enable CI Scripts to Run Locally

ajschmidt8:
This PR introduces some changes to our CI scripts that enable them to be run locally.

In a nutshell, this PR effectively just adds some prompts that allow users to provide values for environment variables that are typically provided by GitHub Actions.

These environment variables ensure that the build and test scripts run successfully outside of a GH Actions environment.

Breaking down the major changes in this PR by file:

- `tools/rapids-download-conda-from-s3`:
  -  the changes in this file prompt the user for the necessary env vars that are needed to identify which artifacts should be downloaded from downloads.rapids.ai
- `tools/rapids-download-from-s3`:
  - the changes in this file ensure that artifacts are obtained from downloads.rapids.ai (instead of via the `aws` CLI) when not in a CI environment. VPN access is required to download these artifacts
- `tools/rapids-env-update`:
  - sets `sccache` to use read-only mode (see [here](https://github.com/mozilla/sccache/blob/412f7e00687f388b7fbe14543e307090db558c64/docs/S3.md)) since proper write credentials aren't available to devs running local builds. sets sensible defaults for other build related vars.
- `tools/rapids-upload-conda-to-s3`:
  - exits early before any uploads are attempted since upload should only happen in GH Actions

The result of these changes is that devs can now volume mount their locally checked out repository into our CI containers and effortlessly run build / test scripts the same way CI does:

```sh
docker run \
  --rm \
  -it \
  --gpus all \
  -v $PWD:/cugraph -w /cugraph \
  rapidsai/ci:cuda11.4.1-ubuntu18.04-py3.8

## then inside the container...

./ci/test_cpp.sh
```

This will help developers easily reproduce any build/test failure that they might encounter in CI with conda packages.

### DIFF
--- a/tools/rapids-constants
+++ b/tools/rapids-constants
@@ -1,3 +1,4 @@
 #!/bin/bash
 
 export RAPIDS_DOWNLOADS_BUCKET="rapids-downloads"
+export RAPIDS_DOWNLOADS_DOMAIN="downloads.rapids.ai"
--- a/tools/rapids-download-conda-from-s3
+++ b/tools/rapids-download-conda-from-s3
@@ -4,7 +4,8 @@
 # Positional Arguments:
 #   1) a string of "cpp" or "python" which determines which conda artifact
 #      should be downloaded
-set -e
+set -euo pipefail
+export RAPIDS_SCRIPT_NAME="rapids-download-conda-from-s3"
 
 pkg_type="$1"
 case "${pkg_type}" in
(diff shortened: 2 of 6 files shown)


ajschmidt8:
Proper docs for devs will follow.

sevagh (approved):

ajschmidt8:
Properly documented in https://github.com/rapidsai/docs/pull/335

PR #39, opened 2023-01-26
Add `DATE_STRING` environment variable for builds

ajschmidt8:
This PR adds the `DATE_STRING` environment variable which will be used in some upcoming conda recipe changes.

Assuming the changes work as intended, we should remove the conditional block of code and remove `VERSION_SUFFIX` in favor of `DATE_STRING` after the `23.02` release.


ajschmidt8:
for reference, the `export` statement is on its own line to account for this shellcheck rule about dynamically computed variables: https://www.shellcheck.net/wiki/SC2155

raydouglass (approved):

PR #40, opened 2023-02-01
Use commit for date string & update var name

ajschmidt8:
@bdice pointed out that it's probably a better idea to get the date from the commit rather than arbitrarily calling the `date` function.

This PR adjusts that implementation accordingly.

It also updates the variable name from `DATE_STRING` to `RAPIDS_DATE_STRING` in accordance with our environment variable name conventions.

### DIFF
--- a/tools/rapids-env-update
+++ b/tools/rapids-env-update
@@ -20,6 +20,6 @@ fi
 
 # TODO: Remove this conditional block after 23.02 release
 if ! rapids-is-release-build; then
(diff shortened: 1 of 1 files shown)


bdice (approved):
To add to the description: this is helpful because it means that package builds will be reproducible. The date of the package will match the commit date, and not the date that CI was run. Re-running CI (or having multiple CI jobs in a workflow that cross an arbitrary midnight boundary) will result in the same date string.

Ethyling (approved):

PR #48, opened 2023-03-01
Clean cache after `conda` segfault

ajschmidt8:
Based on the logs in the run below, it seems that a retry alone is not enough to resolve the segfault issues that have been appearing.

- https://github.com/rapidsai/cudf/actions/runs/4298659489/jobs/7493440631#step:6:403

We should also try cleaning the cache based on the `invalid tarball` messages just under the log line linked above.


ajschmidt8:
admin merging since this change is benign and the error message has been popping up a lot lately.

PR #51, opened 2023-04-25
Add scripts for assembling docker multiarch manifests from a local registry

sevagh:
This PR introduces 3 new gha-tools intended for building docker multiarch images from native hardware:
* `rapids-upload-docker-to-s3`, which takes a local docker tag and uploads it as a tarball to downloads.rapids.ai
* `rapids-download-docker-from-s3`, which downloads a docker tag from downloads.rapids.ai and loads it
* `rapids-docker-multiarch-from-s3`, which downloads all of the uploaded docker tarballs, loads them, pushes them to an ephemeral docker registry, assembles a multiarch manifest, and pushes that to the same ephemeral registry

Currently, building a multiarch image from two local separate architecture images is difficult/impossible: https://github.com/docker/cli/issues/3350 

In cibuildwheel-imgs, I came up with an approach as follows:

For the purposes of this written description, let's assume we have a multiarch image `foo`, for whom we want each of x86_64 and aarch64 to be built natively on a host of that same architecture

1. First, on native hosts (amd64 and arm64), build each image, appending the runner arch: `foo-x86_64`, `foo-aarch64`
2. Use `rapids-upload-docker-to-s3 foo-$arch` to upload `foo-$arch` tarballs to downloads.rapids.ai from each native builder instance
3. [**OPTIONAL**] On native hosts, use `rapids-download-docker-from-s3` to fetch the image, run security scans, unit tests, GPU tests, etc., as needed, before moving onto the final layer
4. In the final layer, the "multiarch assembler", do the following
    1. Set up an ephemeral remote docker registry during the GHA workflow run, using the `service:` key: https://github.com/rapidsai/cibuildwheel-imgs/blob/3943c3f7f22fe7e578dce334a24885fdd8e5e432/.github/workflows/docker-multiarch-native.yml#L109 
        1. This registry is an insecure local registry running as `localhost:5000`
    2. Use `rapids-docker-multiarch-from-s3`, which will do this:
        1. Download `foo-x86_64` from downloads.rapids.ai and `docker load` it
        2. Download `foo-aarch64` from downloads.rapids.ai and `docker load` it
        5. Prepend `localhost:5000/foo-x86_64` and push it to the local registry
        6. Prepend `localhost:5000/foo-aarch64` and push it to the local registry
        7. Now that these are available in a "remote" registry, we can multiarch assemble it
        8. Use `docker manifest create localhost:5000/foo --amend localhost:5000/foo-x86_64 --amend localhost:5000/foo-aarch64` and push it
    3. At this point, `localhost:5000/` contains `foo`, `foo-aarch64`, and `foo-x86_64`
    4. The final step is `skopeo copy localhost:5000/foo dockerhub/foo`
        1. We did all the remote pushing mess in an ephemeral registry, and copy only the final result to dockerhub

Note that the `rapids-download-docker-from-s3` can be used to download each per-arch image in an extra step, for example to do a security scan or unit tests.

### DIFF
--- a/tools/rapids-docker-multiarch-from-s3
+++ b/tools/rapids-docker-multiarch-from-s3
@@ -62,7 +62,7 @@ for OBJ in $(jq -nr 'env.DOCKER_TARBALLS | fromjson | .[]'); do
   MULTIARCH_IMAGES["${multiarch_tag}"]+=" ${loaded_image}"
 done
 
-manifests_to_push=()
+manifests_to_copy=()
 
 for key in "${!MULTIARCH_IMAGES[@]}"; do
   values="${MULTIARCH_IMAGES[$key]}"
@@ -77,16 +77,22 @@ for key in "${!MULTIARCH_IMAGES[@]}"; do
     manifest_args+="--amend ${local_name} "
   done
 
+  local_manifest="localhost:5000/${key}"
+
   # shellcheck disable=SC2086
-  docker manifest create --insecure "${key}" ${manifest_args} 1>&2
-  manifests_to_push+=("${key}")
+  docker manifest create --insecure "${local_manifest}" ${manifest_args} 1>&2
+  docker manifest push "${local_manifest}"
+  manifests_to_copy+=("${local_manifest}")
 
   # if latest tags have been supplied this image matches the latest, give it the additional latest tag
   if [ -n "${latest_tag}" ] && [ -n "${latest_tag_alias}" ] && [ "${key}" == "${latest_tag}" ]; then
+    local_latest="localhost:5000/${latest_tag_alias}"
+
     # shellcheck disable=SC2086
-    docker manifest create --insecure "${latest_tag_alias}" ${manifest_args} 1>&2
-    manifests_to_push+=("${latest_tag_alias}")
+    docker manifest create --insecure "${local_latest}" ${manifest_args} 1>&2
+    docker manifest push "${local_latest}"
+    manifests_to_copy+=("${local_latest}")
   fi
 done
 
-echo -n "${manifests_to_push[@]}"
+echo -n "${manifests_to_copy[@]}"
--- a/tools/rapids-docker-multiarch-from-s3
+++ b/tools/rapids-docker-multiarch-from-s3
@@ -1,0 +1,55 @@
+#!/bin/bash
+# A utility script that downloads all docker images from S3
+# and assembles them into multiarch tags
+set -eo pipefail
+source rapids-constants
+export RAPIDS_SCRIPT_NAME="rapids-docker-multiarch-from-s3"
+
+DOCKER_TMP_DIR="$(mktemp -d)"
+
+DOCKER_STARTS_WITH="docker_"
+DOCKER_ENDS_WITH=".tar.gz"
+
+S3_PATH=$(rapids-s3-path)
+BUCKET_PREFIX=${S3_PATH/s3:\/\/${RAPIDS_DOWNLOADS_BUCKET}\//} # removes s3://rapids-downloads/ from s3://rapids-downloads/ci/rmm/...
+
+# shellcheck disable=SC2016
+DOCKER_TARBALLS=$(
+  set -eo pipefail;
+  aws \
+    --output json \
+    s3api list-objects \
+    --bucket "${RAPIDS_DOWNLOADS_BUCKET}" \
+    --prefix "${BUCKET_PREFIX}" \
+    --page-size 100 \
+    --query "Contents[?starts_with(Key, '${DOCKER_STARTS_WITH}') && ends_with(Key, '${DOCKER_ENDS_WITH}')].Key" \
+    | jq -c
+)
+export DOCKER_TARBALLS
+
+# create an associative array (i.e. dict) of multiarch image to per-arch image
+declare -A MULTIARCH_IMAGES
+
+# download and load them all
+for OBJ in $(jq -nr 'env.DOCKER_TARBALLS | fromjson | .[]'); do
+  FILENAME=$(basename "${OBJ}")
+  S3_URI="${S3_PATH}${FILENAME}"
+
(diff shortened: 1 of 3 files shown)


tools/rapids-docker-multiarch-from-s3
vyasr, line 1: I notice a lot of output redirection. Should we consider placing the entire file inside a group and redirecting? Something like 
```
{
... # The current script contents
} 1>&2]
would work IIRC
```

tools/rapids-docker-multiarch-from-s3
```diff
@@ -0,0 +1,101 @@
...
+manifests_to_copy=()
+
+for key in "${!MULTIARCH_IMAGES[@]}"; do
+  values="${MULTIARCH_IMAGES[$key]}"
+  rapids-echo-stderr "Preparing multiarch manifest for: ${key} with per-arch images: ${values}"
+
+  manifest_args=""
+  for value in ${values}; do
+    # use the local registry set up by the `service` block in GHA
+    local_name="localhost:5000/${value}"
```
vyasr, line 77: Where does the port number come from? Could we avoid hardcoding it here? If we have to hardcode it in the script, I see it used in multiple places so at least a single variable would be helpful.

tools/rapids-docker-multiarch-from-s3
```diff
@@ -0,0 +1,101 @@
...
+    docker tag "${value}" "${local_name}" 1>&2
+    docker push "${local_name}" 1>&2
+    manifest_args+="--amend ${local_name} "
+  done
+
+  local_manifest="localhost:5000/${key}"
```
vyasr, line 83: localhost

tools/rapids-docker-multiarch-from-s3
```diff
@@ -0,0 +1,101 @@
...
+
+  # shellcheck disable=SC2086
+  docker manifest create --insecure "${local_manifest}" ${manifest_args} 1>&2
+  docker manifest push "${local_manifest}" 1>&2
+  manifests_to_copy+=("${key}")
+
+  # if latest tags have been supplied this image matches the latest, give it the additional latest tag
+  if [ -n "${latest_tag}" ] && [ -n "${latest_tag_alias}" ] && [ "${key}" == "${latest_tag}" ]; then
+    local_latest="localhost:5000/${latest_tag_alias}"
```
vyasr, line 92: localhost

tools/rapids-download-docker-from-s3
vyasr, line 1: Should this script be written to reuse `rapids-download-from-s3`?

tools/rapids-download-docker-from-s3
```diff
@@ -0,0 +1,28 @@
...
+
+docker_image="$1"
+
+# replace "/..." with "_..." to use as file name
+docker_image_no_slash="${docker_image//\//_}"
+docker_image_s3_name="docker_${docker_image_no_slash}.tar.gz"
+tmp_fname="${tmp_dir}/${docker_image_no_slash}.tar.gz"
+
+# download .tar.gz into tmpdir
+s3_dl_path="$(rapids-s3-path)${docker_image_s3_name}"
```
vyasr, line 24: Hmm I guess this bit is different from `rapids-download-from-s3`. Also `rapids-download-from-s3` assumes that the uploaded content is a tarball. In this case I guess we don't want to untar it because `docker load` accepts a `tar.gz` directly? Should we rework `rapids-download-from-s3`, or is that too much work? @ajschmidt8 what do you think?

vyasr (changes requested):
Few small questions

tools/rapids-docker-multiarch-from-s3
sevagh: The redirecting is so that the final output to stdout is the list of images (to be used in a `skopeo copy` loop in the push step). I'm ok with all of the previous parts of the script until the final `echo -n` to accommodate your suggestion.

tools/rapids-docker-multiarch-from-s3
sevagh: It's the default port of the docker registry image (launched in the containing workflow). Agreed that it makes less sense for this script to contain a hardcoded value that only makes sense in the outer workflow.

If this were to be transformed into a variable, it could be either an argument or env var, something like `INTERMEDIATE_DOCKER_REGISTRY` or `LOCAL_DOCKER_REGISTRY`?

tools/rapids-download-docker-from-s3
sevagh: They're subtly different; this one uses the s3api to locate all tarballs produced by the current workflow run (with the given sha/repo/branch/etc.).

tools/rapids-download-docker-from-s3
sevagh: `rapids-download-from-s3` could be stale; I never used it much and it might be in an older/less working state. But, it would still be useful if we were to ever run a separate unit test or security scan layer (and want to download the intermediate containers to test/scan).

tools/rapids-docker-multiarch-from-s3
vyasr: I like `LOCAL_DOCKER_REGISTRY`.

vyasr: Ideally with a comment indicating where this requirement is coming from.

tools/rapids-download-docker-from-s3
vyasr: Gotcha. If it's not worth trying to combine them, then perhaps add a comment stating why?

tools/rapids-download-docker-from-s3
vyasr: I don't follow here. It seems like any script has an equal chance of going stale, wouldn't we be better off consolidating so that we have a single point of maintenance and know not to let it go stale?

vyasr (approved):
Approving, will let @ajschmidt8 decide what to do with my outstanding comments when he reviews and makes a final call.

ajschmidt8 (approved):
these scripts aren't used very widely yet.

i'm fine with merging them as-is and iterating as necessary in the future.

PR #53, opened 2023-05-09
Update `RAPIDS_DATE_STRING`

ajschmidt8:
This PR updates how `RAPIDS_DATE_STRING` is computed.

The variable's value is now computed from the workflow run date.

This will help ensure that RAPIDS packages are published on a nightly basis even if no changes to the source code have been made.

### DIFF
--- a/tools/rapids-logger
+++ b/tools/rapids-logger
@@ -20,4 +20,4 @@ function rapids-logger {
   echo -e "\033[32m└${bar}┘\033[0m\n"
 }
 
-rapids-logger "$@"
+rapids-logger "$@" >&2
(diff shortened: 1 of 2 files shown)


tools/rapids-logger
```diff
@@ -20,4 +20,4 @@ function rapids-logger {
   echo -e "\033[32m└${bar}┘\033[0m\n"
 }
 
-rapids-logger "$@"
+rapids-logger "$@" >&2
```
ajschmidt8, line 23: This change is necessary to ensure that the command's logging output is not captured by variables when using `rapids-retry` like this:

```sh
SOME_VAR=$(rapids-retry ...)
```

raydouglass (approved):

PR #55, opened 2023-05-18
wheel-ctk-name-gen

divyegala:
Part of https://github.com/rapidsai/shared-action-workflows/issues/77

### DIFF
--- a/tools/rapids-wheel-ctk-name-gen
+++ b/tools/rapids-wheel-ctk-name-gen
@@ -1,0 +1,18 @@
+#!/bin/bash
+# A utility script that generates CUDA suffix and wheel name for wheel-related workflows
+# Positional Arguments:
+#   1) wheel path to tar up
+set -eu -o pipefail
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+  rapids-echo-stderr "Must specify both input arguments: CTK_TAG and PACKAGE_NAME"
+  exit 1
+fi
+
+ctk_tag="$1"
+ctk_major=$(echo $ctk_tag | cut -d'.' -f1)
+cuda_suffix="-cu${ctk_major}"
+wheelname="$2_cu${ctk_major}"
+
+echo "RAPIDS_PY_WHEEL_CUDA_SUFFIX=${cuda_suffix}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
+echo "RAPIDS_PY_WHEEL_NAME=${wheelname}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"


tools/rapids-wheel-ctk-name-gen
```diff
@@ -0,0 +1,19 @@
...
+  exit 1
+fi
+
+ctk_tag="$1"
+ctk_major=$(echo "$ctk_tag" | cut -d'.' -f1)
+cuda_suffix="-cu${ctk_major}"
+wheelname="$2_cu${ctk_major}"
+
+echo "RAPIDS_PY_WHEEL_CUDA_SUFFIX=${cuda_suffix}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
+echo "RAPIDS_PY_WHEEL_NAME=${wheelname}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
```
wence-, line 19: This is not that useful if someone wants to use this outside the actions workflow.

divyegala: I figured these are just `gha` specific tools. @vyasr do these scripts need to be self contained, and we set actions related variables outside the script?

wence-: If we want to move towards being able to run the steps for doing things locally then it's useful if the tools don't reference github-specific stuff

divyegala: Sounds good, in that case perhaps we should just export these two variables?

tools/rapids-wheel-ctk-name-gen
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# A utility script that generates CUDA suffix
```
vyasr, line 2: Would be good to add small documentation of what the CUDA suffix looks like so that it's clear that we do `cu11` rather than `cuda11`/`cuda_11`/etc.

vyasr (approved):
One small request for additional documentation, otherwise LGTM.

tools/rapids-wheel-ctk-name-gen
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# A utility script that generates CUDA suffix in the format "cu${VER}" where `$VER`
+# is the CUDA version major. (For example, "cu11")
```
vyasr, line 3:
```suggestion
# is the CUDA major version (for example, "cu11").
```

tools/rapids-wheel-ctk-name-gen
```diff
@@ -0,0 +1,17 @@
...
+# Positional Arguments:
+#   1) ctk-tag
```
vyasr, line 5:
```suggestion
#   1) ctk tag
```

vyasr (approved):

PR #57, opened 2023-05-26
Enable wheels CI scripts to run locally

divyegala:
Taking inspiration from #35 and #62

This PR adds / updates:

- `rapids-configure-sccache` (new) : configures sccache for CI/local, with the intention being to remove this information from CI containers
- `rapids-date-string` (new) : configures date string
- `rapids-configure-conda-channels` (new) : modifies conda channels based on build type
- `rapids-env-update` (modified) : calls the `rapids-configure-sccache`
- `rapids-prompt-local-repo-config` (new) : consolidates prompts for users to configure repo information locally 
- `rapids-download-wheels/conda-from-s3` (updated) : uses `rapids-prompt-local-repo-config` 
- `rapids-upload-wheels-to-s3` (modifed) : guards to ensure local builds don't try to upload


vyasr:
Is there a way to avoid duplicating this logic from the conda builds? Maybe putting it into a separate tool? I agree that we don't want to (nor can we really) lump the logic into rapids-package-name, but I also would like to share conda/wheel logic on this.

vyasr:
Also you may want to make sure that whatever you do supports https://github.com/rapidsai/gha-tools/pull/62/ (another reason to avoid duplicating if possible).

divyegala:
I already copied over #62 here.

I'll think on how to minimize duplication, maybe we can combine the prompts as part of another tool.

vyasr:
Yeah another tool seems like the lowest hanging fruit.

divyegala:
@ajschmidt8 as we move towards local builds, @vyasr and I discussed that some steps may be entirely skippable locally and the build would still be able to succeed. Uploading to S3 is one such step, and I don't think we should exit early.

tools/rapids-configure-sccache
```diff
@@ -0,0 +1,19 @@
...
+  export PARALLEL_LEVEL=${PARALLEL_LEVEL:-$(nproc)}
+  export RAPIDS_BUILD_TYPE=${RAPIDS_BUILD_TYPE:-"pull-request"}
+else
+  export CMAKE_CUDA_COMPILER_LAUNCHER=sccache
+  export CMAKE_CXX_COMPILER_LAUNCHER=sccache
+  export CMAKE_C_COMPILER_LAUNCHER=sccache
+  export SCCACHE_BUCKET=rapids-sccache-east
+  export SCCACHE_REGION=us-east-2
+  export SCCACHE_IDLE_TIMEOUT=32768
+  export SCCACHE_S3_USE_SSL=true
```
vyasr, line 17: We want all of these set unconditionally. The only piece that's specific to local builds is specifying that we don't have write credentials AFAIK. @ajschmidt8 can you confirm?

tools/rapids-configure-sccache
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# A utility script that configures sccache environment variables
+
+if [ "${CI:-false}" = "false" ]; then
+  # Configure sccache for read-only mode since no credentials
+  # are available in local builds.
+  export SCCACHE_S3_NO_CREDENTIALS=true
+  export PARALLEL_LEVEL=${PARALLEL_LEVEL:-$(nproc)}
+  export RAPIDS_BUILD_TYPE=${RAPIDS_BUILD_TYPE:-"pull-request"}
```
vyasr, line 9: These don't seem sccache-specific. @ajschmidt8 do you recall why these needed to be set specifically in the CI=false case? That would help indicate where the proper place to set these is.

tools/rapids-prompt-local-repo-config
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+# A utility script that prompts user to setup repository information in
+# local environments
+
+if [ "${CI:-false}" = "false" ]; then
+  rapids-echo-stderr "Local run detected."
+  rapids-echo-stderr "NVIDIA VPN connectivity is required to download workflow artifacts."
+
+  if [ -z "${RAPIDS_BUILD_TYPE:-}" ]; then
```
vyasr, line 9: Is there any point in the `:-` if there's no default afterwards?

vyasr: Same question applies to the other vars here as well.

tools/rapids-configure-conda-channels
vyasr, line 1: I don't mind having this as a separate script, but FWIW this could stay inlined couldn't it? There's no reuse so splitting it out seems premature. NBD though, I don't mind this change just asking.

tools/rapids-configure-sccache
vyasr, line 1: I guess because of the fact that this is setting environment variables we're going to have to source this script in every individual build_wheel*.sh script, right? That's a bit annoying. The other option would be to call this in the build-wheels.yaml workflow, and then manually reexport all the variables to `GITHUB_ENV`. Both options introduce duplication. Which do we prefer?

The result doesn't change this PR, but trying to plan ahead here.

tools/rapids-date-string
vyasr, line 1: This is not currently reused but will eventually be reused to replace the epoch timestamp for wheel builds, right?

vyasr (changes requested):

tools/rapids-date-string
divyegala: Yes, it's also called in `rapids-env-update`

tools/rapids-configure-sccache
divyegala: Prefer the former because the idea is to ensure write access is denied for local builds. Exporting anything from `GITHUB_ENV` makes the process not reproducible locally

tools/rapids-configure-conda-channels
divyegala: We'll eventually remove `rapids-env-update` from conda build scripts as well, and call these new, smaller tools there

tools/rapids-configure-sccache
divyegala: I can at least remove `RAPIDS_BUILD_TYPE`, the prompts already take care of it

tools/rapids-prompt-local-repo-config
divyegala: I suppose it initializes the variable to an empty string? Not sure if it's necessary

tools/rapids-configure-sccache
vyasr: Oh of course good point. OK yeah so not really much that we can do about that duplication then.

tools/rapids-date-string
vyasr: Right, it's only used once (in rapids-env-update) right now is all I meant.

tools/rapids-prompt-local-repo-config
ajschmidt8: This is necessary because of the `set -u` option that we use at the top of each script.

`set -u` treats unset variables as errors ([src](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)).

If we didn't include the `:-` syntax in `"${RAPIDS_BUILD_TYPE:-}"`, the script would error that `RAPIDS_BUILD_TYPE` is not set.

Using `:-` to initialize it as an empty string if it's not already set allows us to make the conditional statement work with the `set -u` option.

tools/rapids-configure-sccache
ajschmidt8: `PARALLEL_LEVEL` is set as an environment variable on our actual self-hosted runners.

So I figured we should include a value for that variable when running locally.

I'm not sure the best place to put this now that we're splitting up some of these scripts. If there's no objections, it can stay here. Alternatively we can try to find a new script for it or remove it entirely, though I'm not sure of the implications for local builds if we remove it entirely.

It's only one line, so we could also just leave it for now and then move it to the individual repositories in the future.

The `RAPIDS_BUILD_TYPE` line can probably be removed for the reason that Divye mentioned. I was just trying to provide a sane default for people who are trying to build things locally.

tools/rapids-prompt-local-repo-config
```diff
@@ -0,0 +1,77 @@
...
+      echo "Suppress this prompt in the future by setting the 'RAPIDS_REPOSITORY' environment variable:"
+      echo "export RAPIDS_REPOSITORY=${RAPIDS_REPOSITORY}"
+      echo ""
+    } >&2
+  fi
+
+  if [ -z "${RAPIDS_REF_NAME:-}" ]; then
+    {
+      echo ""
+      if [ "${RAPIDS_BUILD_TYPE}" = "pull-request" ]; then
```
ajschmidt8, line 36:
```suggestion
      if [ "${RAPIDS_BUILD_TYPE:-pull-request}" = "pull-request" ]; then
```

If we remove the `RAPIDS_BUILD_TYPE` default value as mentioned below, we should probably re-add it here to preserve the existing functionality.

- https://github.com/rapidsai/gha-tools/pull/57/files/f33c72eb82ef7d8f5bbfe61e0c794fc3d28025d3#r1268473429

tools/rapids-configure-sccache
vyasr: I'm OK leaving `PARALLEL_LEVEL` now and moving it later. That said, since we generally use ninja (and not make) everywhere the default of nproc is the same as what ninja will do anyway.

tools/rapids-prompt-local-repo-config
vyasr: Ah OK got it.

tools/rapids-configure-sccache
divyegala: We can remove it later when we confirm locally that builds work as expected without `PARALLEL_LEVEL` set in the containers

vyasr: Sounds good. Resolving.

tools/rapids-prompt-local-repo-config
vyasr: I agree, I would prefer to remove it altogether.

vyasr (approved):
Everything looks good to me now, pending resolution of AJ's open question regarding RAPIDS_BUILD_TYPE and assuming the gh CLI installation is all good.

tools/rapids-prompt-local-repo-config
ajschmidt8: I'm fine with this :+1:

ajschmidt8 (approved):

PR #59, opened 2023-06-08
Make `rapids-get-artifact` work locally

ajschmidt8:
This PR ensures that `rapids-get-artifact` works locally.

This request was made by @bdice who was attempting to use the script locally to debug some CUDA 12 issues.

I've moved the logic for downloading, extracting, and printing the artifact's path to a new script, `_rapids-download-from-s3`.

This script is then called by both `rapids-download-from-s3` and `rapids-get-artifact`.

In the future, I would like to rework some of the scripts regarding artifact uploading/downloading, but I think we need to finish reworking the wheel workflows before that will be feasible.

### DIFF
--- a/tools/_rapids-download-from-s3
+++ b/tools/_rapids-download-from-s3
@@ -1,0 +1,27 @@
+#!/bin/bash
+# A utility script that downloads an artifact archive from S3, untars it,
+# and prints the location where it was untarred.
+# The script can be used locally and in CI.
+# Positional Arguments:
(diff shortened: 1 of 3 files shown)


ajschmidt8:
I've tested these changes locally with the process outlined in #60 and everything seems to work correctly.

AyodeAwe (approved):

raydouglass (approved):

bdice (approved):
This looks good to me.

PR #68, opened 2023-08-01
Fix `rapids-date-string`

ajschmidt8:
As it turns out, using the `YYMMDD` format in `rapids-date-string` for wheels broke some things.

This PR adjusts `rapids-date-string` to generate both `YYMMDD` strings and "seconds since epoch" strings for wheels.

Since the new wheel scripts all call `rapida-date-string` without any arguments, the default argument is set to "second".

The default argument should be removed once `rapids-env-update` is deprecated and the individual repositories have been updated to call `rapids-date-string` in this new format.

`rapids-env-update` has also been updated to pass the `day` argument since all conda workflows currently get their `RAPIDS_DATE_STRING` value through that script.

To avoid having to pass arguments from the calling scripts in the future, we could create two new `gha-tools` scripts (e.g. `rapids-wheel-date-string` and `rapids-conda-date-string`) that call a "private" script (`_rapids-date-string`) with the appropriate arguments.

### DIFF
--- a/tools/rapids-date-string
+++ b/tools/rapids-date-string
@@ -1,9 +1,25 @@
 #!/bin/bash
 # A utility script produces a date string
 
-RAPIDS_DATE_STRING="$(date +%y%m%d)"
+# TODO: remove default value after calling scripts are updated
+FORMAT_ARG="${1:-second}"
+
+case "${FORMAT_ARG}" in
+  day)
+    FORMAT="%y%m%d"
+    ;;
+  second)
+    FORMAT="%s"
+    ;;
+  *)
+    rapids-echo-stderr "Invalid format argument: ${FORMAT_ARG}"
+    exit 1
+    ;;
+esac
+
+RAPIDS_DATE_STRING="$(date +${FORMAT})"
 if [ "${GITHUB_ACTIONS:-false}" = "true" ]; then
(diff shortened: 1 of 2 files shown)


ajschmidt8:
the format string for wheels was taken from here: https://github.com/rapidsai/shared-action-workflows/blob/6eff70b1f7614a0d16c9f499280d0106be3c0fd7/.github/workflows/wheels-pure-build.yml#L61

ajschmidt8:
You can test this out locally with the following commands (run from the root of the repo):

```sh
# invalid
(PATH=$PATH:$PWD/tools source ./tools/rapids-date-string abcdef; echo $RAPIDS_DATE_STRING)

# day
(PATH=$PATH:$PWD/tools source ./tools/rapids-date-string day; echo $RAPIDS_DATE_STRING)

# second
(PATH=$PATH:$PWD/tools source ./tools/rapids-date-string second; echo $RAPIDS_DATE_STRING)

# no arg
(PATH=$PATH:$PWD/tools source ./tools/rapids-date-string; echo $RAPIDS_DATE_STRING)
```

vyasr (approved):
This change looks fine for now. I think we'll want to revisit the nightly versioning scheme for wheels in general. That will inform any final form of rapids-date-string for wheels since we may not want to embed a date or timestamp in the way that we currently are.

PR #71, opened 2023-08-02
Remove rapids_twine_new

vyasr:
This tool is now completely redundant, rapids_twine can always be used instead.


vyasr:
Since we don't follow the same branching strategy in gha-tools as the rest of RAPIDS, it's probably best to wait to merge this until after we've completed 23.08 release builds so that this doesn't accidentally get into images used to build 23.08 packages that still use workflows that use this tool.

bdice (approved):

ajschmidt8 (approved):

PR #75, opened 2023-09-08
Remove path constraint from PR workflow

AyodeAwe:
The constraint prevented checks from being run on PRs


ajschmidt8 (approved):
Approving, pending a PR description being added.

PR #79, opened 2023-09-28
Include additional docs upload format "xml_tar"

AyodeAwe:
In the ongoing work to consolidate the rendered docs for all graph repositories (`cugraph`, `cugraph-ops`, `wholegraph`) into one place, `wholegraph` and `cugraph-ops` c++ docs will be uploaded as xml tarballs, which will then be consumed during the docs build process for `cugraph` and ultimately rendered under `cugraph`.

This PR enables the tarball upload.

### DIFF
--- a/tools/rapids-upload-docs
+++ b/tools/rapids-upload-docs
@@ -77,7 +77,7 @@ copy_docs_to_s3() {
       fi
 
       if [[ "${FORMAT}" == "xml_tar" ]]; then
-        NUM_FILES=$(find "$PROJECT_FORMAT_DIR" -maxdepth 1 -type f | wc -l)
+        NUM_FILES=$(find "$PROJECT_FORMAT_DIR" -type f | wc -l)
         if [[ ! -f "${PROJECT_FORMAT_DIR}/xml.tar.gz" || $NUM_FILES -ne 1 ]]; then
             echo "Error: The xml_tar directory should contain a single file named xml.tar.gz."
             exit 1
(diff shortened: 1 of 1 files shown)


tools/rapids-upload-docs
```diff
@@ -71,16 +71,30 @@ copy_docs_to_s3() {
...
 
-      if [[ ! "${FORMAT}" =~ ^(html|txt)$ ]]; then
-        echo "ERROR: FORMAT must be either 'html' or 'txt'."
+      if [[ ! "${FORMAT}" =~ ^(html|txt|xml_tar)$ ]]; then
+        echo "ERROR: FORMAT must be either 'html' or 'txt' or 'xml_tar'."
         exit 1
       fi
 
+      if [[ "${FORMAT}" == "xml_tar" ]]; then
+        NUM_FILES=$(find "$PROJECT_FORMAT_DIR" -maxdepth 1 -type f | wc -l)
```
ajschmidt8, line 80:
```suggestion
        NUM_FILES=$(find "$PROJECT_FORMAT_DIR" -type f | wc -l)
```

Do we need `-maxdepth`? seems safer to omit it in case there are subdirectories (which we don't want)

tools/rapids-upload-docs
```diff
@@ -71,16 +71,30 @@ copy_docs_to_s3() {
...
 
+      if [[ "$FORMAT" == "xml_tar" ]]; then
+          ACL_OPTION="--acl public-read"
+      else
+          ACL_OPTION=""
+      fi
       aws s3 sync \
         --no-progress \
         --delete \
+        "${ACL_OPTION}" \
```
ajschmidt8, line 97:
```suggestion
      ACL_OPTION="private"
      if [[ "$FORMAT" == "xml_tar" ]]; then
          ACL_OPTION="public-read"
      fi
      aws s3 sync \
        --no-progress \
        --delete \
        --acl "${ACL_OPTION}" \
```

Can we ditch the `else` statement in favor of a default value?

Also, I think it's more clear if we include both `private` and `public-read` values.

ajschmidt8 (approved):

PR #80, opened 2023-10-04
Add new tool for generating versions

vyasr:
This tool will be used to generate standardized versions for RAPIDS packages.

### DIFF
--- a/tools/rapids-generate-version
+++ b/tools/rapids-generate-version
@@ -6,7 +6,7 @@
 set -exo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-generate-version"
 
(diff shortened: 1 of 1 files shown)


tools/rapids-generate-version
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# A utility script that generates a package version
+set -exo pipefail
```
ajschmidt8, line 3:
```suggestion
set -euo pipefail
```

this aligns with the settings we use in other scripts.

vyasr: Done.

ajschmidt8 (approved):

PR #82, opened 2023-10-26
Add new github tool that extracts all conda archives

robertmaynard:


ajschmidt8 (approved):

tools/rapids-extract-conda-files
```diff
@@ -0,0 +1,27 @@
...
+
+# dir that the extracted tarball are in
+channel="/tmp/${pkg_type}_channel"
+
+# empty dir to extract files to
+extract="/tmp/${pkg_type}_binaries"
+
+mkdir -p "${extract}"
+cd "${extract}"
+find "${channel}" -name "*tar.bz2" -type f -print0 | xargs -0 -n 1 tar -v -xf
```
ajschmidt8, line 27: Quick request that I missed the first time around.

Can we adapt this script to use `mktemp -d` instead of a hardcoded `/tmp/${pkg_type}_binaries` directory?

Then we can print the name of the directory generated by the `mktemp -d` command at the end of the script so that it can be captured by calling scripts.

This is similar to what we do in other places like [here](https://github.com/rapidsai/cudf/blob/52f7d5c7d5d340c3c9011beaa075babc98c1cc0b/ci/build_python.sh#L14).

To accomplish this, this `rapids-extract-conda-files` script can be edited to look like the following:

```sh
{
  # commands in this block output to stderr
  extract_dir=$(mktemp -d)
  # ... other script commands go in this block
} >&2

# this echoes the dir name for consumption by calling scripts
echo -n "${extract_dir}"
```


This will eliminate the fragility of managing a hardcoded directory in multiple areas.

PR #91, opened 2023-12-14
remove tools for working with S3-hosted docker images

jameslamb:
Proposes removing the following scripts, which appear to be unused.

* `rapids-docker-multiarch-from-s3` ([GitHub search](https://github.com/search?q=org%3Arapidsai%20%22rapids-docker-multiarch-from-s3%22&type=code))
* `rapids-download-docker-from-s3` ([GitHub search](https://github.com/search?q=org%3Arapidsai+%22rapids-download-docker-from-s3%22&type=code))
* `rapids-upload-docker-from-s3` ([GitHub search](https://github.com/search?q=org%3Arapidsai+%22rapids-upload-docker-from-s3%22&type=issues))

Found these while working through https://github.com/rapidsai/ops/issues/2982, looking for how `downloads.rapids.ai` is used.

### Notes for Reviewers

These scripts were introduced in April 2023 (#51).

It seems then that the goal was to remove the need to use DockerHub for intermediate images used in producing the container images that RAPIDS projects run CI in:

* https://github.com/rapidsai/ops/issues/2554#issuecomment-1440511825
* https://github.com/rapidsai/ops/issues/2554#issuecomment-1462204628

Since then https://github.com/rapidsai/cibuildwheel-imgs, the only project these were intended to be used with (as far as I can tell) has been archived (https://github.com/rapidsai/ci-imgs/pull/65).


ajschmidt8 (approved):
agreed! these can definitely be removed.

appreciate the effort here and the thorough write-up :fire: :fire: :fire:

PR #96, opened 2024-03-04
Add support for pure Python wheels.

bdice:
This improves the handling of pure Python wheels in `rapids-package-name`. When dealing with pure Python packages, we don't want the Python version or CPU arch to be a part of the package's identifiers.

After this is merged, we can adopt this change across RAPIDS repos for pure wheels like dask-cudf, cuxfilter, etc. and then revert https://github.com/rapidsai/shared-workflows/pull/188.

List of packages to check:
- [ ] dask-cudf https://github.com/rapidsai/cudf/pull/15223
- [ ] cuxfilter
- [ ] nx-cugraph
- [ ] cugraph-pyg
- [ ] cugraph-dgl
- [ ] cugraph-equivariant
- [ ] rapids-dask-dependency (No -cu suffix)

### DIFF
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -31,11 +31,11 @@ case "${pkg_type}" in
     ;;
   wheel_python)
     append_wheelname=1
-    append_pyver=1
-    append_arch=1
-    ;;
-  wheel_python_pure)
-    append_wheelname=1
+    # Pure wheels do not need a pyver or arch
+    if [[ ! -v RAPIDS_PY_WHEEL_PURE ]] || [[ "${RAPIDS_PY_WHEEL_PURE}" != "1" ]]; then
+      append_pyver=1
+      append_arch=1
+    fi
     ;;
   *)
     rapids-echo-stderr "Nonstandard package type '${pkg_type}'"
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -57,7 +57,7 @@ fi
 
 # for python package types (except pure wheels), add pyver
 if (( append_pyver == 1 )); then
-  pkg_name+="_${RAPIDS_PY_VERSION//./}"
+  pkg_name+="_py${RAPIDS_PY_VERSION//./}"
 fi
 
 # for cpp and python package types (except pure wheels), append the arch
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -15,19 +15,26 @@ fi
 pkg_type="$1"
 
 append_cuda=0
-append_pyver=0
 append_wheelname=0
+append_pyver=0
+append_arch=0
 
 case "${pkg_type}" in
   conda_cpp)
     append_cuda=1
+    append_arch=1
     ;;
   conda_python)
     append_cuda=1
     append_pyver=1
+    append_arch=1
     ;;
   wheel_python)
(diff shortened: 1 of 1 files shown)


tools/rapids-package-name
```diff
@@ -48,21 +55,23 @@ if (( append_wheelname )) && [[ -v RAPIDS_PY_WHEEL_NAME ]] && [[ "${RAPIDS_PY_WH
   pkg_name+="_${RAPIDS_PY_WHEEL_NAME}"
 fi
 
-# for python package types, add pyver
+# for python package types (except pure wheels), add pyver
 if (( append_pyver == 1 )); then
   pkg_name+="_${RAPIDS_PY_VERSION//./}"
```
bdice, line 60: Could we also modify this to include "py"? This was requested recently by @jakirkham and seems like a reasonable change to me.
```suggestion
  pkg_name+="_py${RAPIDS_PY_VERSION//./}"
```

tools/rapids-package-name
```diff
@@ -15,19 +15,26 @@ fi
 pkg_type="$1"
 
 append_cuda=0
-append_pyver=0
 append_wheelname=0
+append_pyver=0
+append_arch=0
```
vyasr, line 20: I'm surprised you didn't take this opportunity to alphabetize these :slightly_smiling_face:

tools/rapids-package-name
vyasr: Yeah that is fine with me.

vyasr (approved):
Looks fine to me. Once this is in we'll need follow-ups in the relevant repos.

tools/rapids-package-name
bdice: I'm sorting it by the order in which they appear in the filename.

bdice:
https://github.com/rapidsai/cudf/pull/15223 succeeded and all looks fine.

@vyasr Do you want to re-review? I made some substantial changes since you approved. We should be good to merge this with one review.

tools/rapids-package-name
```diff
@@ -15,20 +15,27 @@ fi
...
   conda_python)
     append_cuda=1
     append_pyver=1
+    append_arch=1
     ;;
   wheel_python)
-    append_pyver=1
     append_wheelname=1
+    # Pure wheels do not need a pyver or arch
+    if [[ ! -v RAPIDS_PY_WHEEL_PURE ]] || [[ "${RAPIDS_PY_WHEEL_PURE}" != "1" ]]; then
```
vyasr, line 35: Honestly probably more sophisticated handling than we strictly need. I'd also be fine with just the first clause and skipping the second check. Also, did someone request the change from a different argument to an env var?

I don't have strong feelings on either so feel free to merge as-is, just wondering what prompted those changes.

vyasr (approved):

tools/rapids-package-name
bdice: This script is not called directly. You have to inject the knowledge of whether the wheel is pure or not when calling the higher level wheel upload/download script. This is the same way that we are injecting the custom wheel name for some packages so I copied the design of using an env var as input.

vyasr: Ah right sorry I got turned around because of how you did this originally.

PR #104, opened 2024-04-25
make rapids-configure-conda-channels idempotent

jameslamb:
While testing some RAPIDS projects, I've found it helpful to do something like this to test `conda builds`.

```shell
docker run \
    --rm \
    -v $(pwd):/opt/rmm \
    -w /opt/rmm \
    -it rapidsai/ci-conda:cuda12.2.2-ubuntu22.04-py3.9-amd64 \
    bash

# build a conda package
ci/build_cpp.sh
```

*(example: https://github.com/rapidsai/build-planning/issues/54#issuecomment-2072893399)*

In that interactive case, it's useful to be able to modify the source of the project and directly run that `ci/build_*` script again.

Many of the RAPIDS projects' build scripts include a call to `rapids-configure-conda-channels` near the beginning. ([GitHub search](https://github.com/search?q=org%3Arapidsai+%22rapids-configure-conda-channels%22+path%3Aci%2Fbuild_cpp.sh&type=code). That interrupts this workflow because that script is not idempotent... it raises an exception if asked to remove a channel that isn't currently in the conda configuration.

```text
CondaKeyError: 'channels': 'rapidsai' is not in the 'channels' key of the config file
```

This proposes changing the behavior of that tool:

- **before**: *try to remove `{channel}` from the conda configuration, raise an error if `{channel}` isn't present there*
- **after**: *if `{channel}` exists in the conda configuration, remove it, otherwise warn that it already doesn't exist there*

### Notes for Reviewers

I can think of 2 reasons we might decide not to merge this:

* having this be a loud error means typos will be caught at CI time... e.g. trying to remove a channel called `rapidz` (note the `z`) would result in a CI failure on `main` but just a log message on this branch
* this branch's changes make a fairly simple script more complex, which makes it harder to understand

I think these risks are worthwhile in exchange for removing some friction from local development on RAPIDS projects. I'm especially not too worried about the typo problem, because the channels this tool references are string literals within the tool, not coming through other configuration that's customized by projects.

### How I tested this

<details><summary>in a container, using one of the images we use when building wheels in CI (click me)</summary>

From the root of this repo, on this branch.

```shell
docker run \
    --rm \
    -v $(pwd):/opt/gha-tools \
    -w /opt/gha-tools \
    -it rapidsai/ci-conda:cuda12.2.2-ubuntu22.04-py3.9-amd64 \
    bash
```

Checked the current set of configured channels.

```shell
conda config --get channels
```

That printed the following.

```text
--add channels 'nvidia'   # lowest priority
--add channels 'conda-forge'
--add channels 'pytorch'
--add channels 'dask/label/dev'
--add channels 'rapidsai-nightly'
--add channels 'rapidsai'   # highest priority
```

Ran the script.

```shell
./tools/rapids-configure-conda-channels
```

Saw this in its logs

```text
[rapids-is-release-build] is not release build
```

Checked the channels again.

```shell
conda config --get channels
```

As expected, `rapidsai` was removed.

```text
--add channels 'nvidia'   # lowest priority
--add channels 'conda-forge'
--add channels 'pytorch'
--add channels 'dask/label/dev'
--add channels 'rapidsai-nightly'   # highest priority
```

Ran the script another 5 times, just to prove that it was really idempotent.

```shell
for i in {1..5}; do
    ./tools/rapids-configure-conda-channels
done
```

All of those succeeded, and the expected warning was issued.

```text
[rapids-is-release-build] is not release build
[rapids-configure-conda-channels] channel 'rapidsai' not found via 'conda config --get channels'
[rapids-is-release-build] is not release build
[rapids-configure-conda-channels] channel 'rapidsai' not found via 'conda config --get channels'
[rapids-is-release-build] is not release build
[rapids-configure-conda-channels] channel 'rapidsai' not found via 'conda config --get channels'
[rapids-is-release-build] is not release build
[rapids-configure-conda-channels] channel 'rapidsai' not found via 'conda config --get channels'
[rapids-is-release-build] is not release build
[rapids-configure-conda-channels] channel 'rapidsai' not found via 'conda config --get channels'
```

Repeating this same process on `main`, the second call fails like this:

```text
CondaKeyError: 'channels': 'rapidsai' is not in the 'channels' key of the config file
```

</details>

### DIFF
--- a/tools/rapids-configure-conda-channels
+++ b/tools/rapids-configure-conda-channels
@@ -10,7 +10,7 @@ conda_channel_in_config() {
 }
 
 # Only try to run 'conda config --remove' if the channel exists in the config.
-# This is here to avoid errors if this script is invoked multiple times in the same
+# This is here to avoid errors if this script is invoked multiple times in the same environment.
 remove_conda_channel() {
   channel_id=${1:?err}
   in_config=$(conda_channel_in_config "${channel_id}")
--- a/tools/rapids-configure-conda-channels
+++ b/tools/rapids-configure-conda-channels
@@ -13,7 +13,6 @@ conda_channel_in_config() {
 # This is here to avoid errors if this script is invoked multiple times in the same environment.
 remove_conda_channel() {
   channel_id=${1:?err}
-  in_config=$(conda_channel_in_config "${channel_id}")
   if conda_channel_in_config "${channel_id}" > /dev/null; then
     conda config --system --remove channels "${channel_id}"
   else
--- a/tools/rapids-configure-conda-channels
+++ b/tools/rapids-configure-conda-channels
@@ -1,11 +1,29 @@
 #!/bin/bash
 # A utility script that configures conda channels
 
+conda_channel_in_config() {
+  channel_id=${1:?err}
+  conda config --json --get channels \
+  | jq -r --arg c "${channel_id}" '."get"."channels" | any(. == $c )'
+}
+
+# Only try to run 'conda config --remove' if the channel exists in the config.
+# This is here to avoid errors if this script is invoked multiple times in the same
+remove_conda_channel() {
+  channel_id=${1:?err}
+  in_config=$(conda_channel_in_config "${channel_id}")
+  if [[ "${in_config}" == "true" ]]; then
+    conda config --system --remove channels "${channel_id}"
+  else
+    echo "[rapids-configure-conda-channels] channel '${channel_id}' not found via 'conda config --get channels'"
+  fi
+}
+
 # Remove nightly channels if build is a release build
 if rapids-is-release-build; then
-  conda config --system --remove channels rapidsai-nightly
-  conda config --system --remove channels dask/label/dev
+  remove_conda_channel 'rapidsai-nightly'
+  remove_conda_channel 'dask/label/dev'
 else
   # exclude stable channel from all non-release builds.
-  conda config --system --remove channels rapidsai
+  remove_conda_channel 'rapidsai'
 fi
--- a/tools/rapids-configure-conda-channels
+++ b/tools/rapids-configure-conda-channels
@@ -1,10 +1,12 @@
 #!/bin/bash
 # A utility script that configures conda channels
 
+set -euo pipefail
+
 conda_channel_in_config() {
   channel_id=${1:?err}
   conda config --json --get channels \
-  | jq -r --arg c "${channel_id}" '."get"."channels" | any(. == $c )'
+  | channel_id="${channel_id}" jq -r --exit-status '.get.channels | any(. == env.channel_id )'
 }
 
 # Only try to run 'conda config --remove' if the channel exists in the config.
@@ -12,7 +14,7 @@ conda_channel_in_config() {
 remove_conda_channel() {
   channel_id=${1:?err}
   in_config=$(conda_channel_in_config "${channel_id}")
-  if [[ "${in_config}" == "true" ]]; then
+  if conda_channel_in_config "${channel_id}" > /dev/null; then
     conda config --system --remove channels "${channel_id}"
   else
     echo "[rapids-configure-conda-channels] channel '${channel_id}' not found via 'conda config --get channels'"
--- a/tools/rapids-check-pr-job-dependencies
+++ b/tools/rapids-check-pr-job-dependencies
@@ -5,21 +5,35 @@
 # This is necessary since the RAPIDS branch protections are configured to require
 # the "pr-builder" job to pass for all PRs. It's implied that that job depends
 # on all other jobs in the workflow.
+#
+# An optional argument can be passed to ignore certain jobs from the check.
+# The argument should be a space-separated list of job names, e.g.:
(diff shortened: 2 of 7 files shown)


tools/rapids-configure-conda-channels
```diff
@@ -1,11 +1,29 @@
...
+
+# Only try to run 'conda config --remove' if the channel exists in the config.
+# This is here to avoid errors if this script is invoked multiple times in the same
+remove_conda_channel() {
+  channel_id=${1:?err}
+  in_config=$(conda_channel_in_config "${channel_id}")
+  if [[ "${in_config}" == "true" ]]; then
+    conda config --system --remove channels "${channel_id}"
+  else
+    echo "[rapids-configure-conda-channels] channel '${channel_id}' not found via 'conda config --get channels'"
```
jameslamb, line 18: Proposing adding this so we have a chance of catching the "there was a typo in the channel name" case.

In theory this should never show up in CI logs unless there's a problem, since this script should only be run once during CI jobs.

tools/rapids-configure-conda-channels
```diff
@@ -1,11 +1,29 @@
...
 # A utility script that configures conda channels
 
+conda_channel_in_config() {
+  channel_id=${1:?err}
+  conda config --json --get channels \
+  | jq -r --arg c "${channel_id}" '."get"."channels" | any(. == $c )'
+}
+
+# Only try to run 'conda config --remove' if the channel exists in the config.
+# This is here to avoid errors if this script is invoked multiple times in the same
```
jameslamb, line 11:
```suggestion
# This is here to avoid errors if this script is invoked multiple times in the same environment.
```

This comment was incomplete.

ajschmidt8 (commented):
What do you think about simplifying all of this into just:

```sh
conda config --system --remove channels rapidsai || true
```

The `|| true` part will prevent it from exiting with a non-zero code if the channel doesn't exist.

And the standard conda error that you shared below should still print:

```
CondaKeyError: 'channels': 'rapidsai' is not in the 'channels' key of the config file
```

jameslamb:
Thanks for reviewing!

I'm proposing not `|| true`-ing it because that'd hide mean these types of issues wouldn't cause an immediate failure:

* `conda config` no longer recognizing the arguments `--system` or `--remove`
* `channels` itself not being a valid key in `conda` config anymore
* `conda` executable not being found on `PATH`

I'll defer to you. If you think the added complexity in the current state of this PR isn't worth the protection against those types of issues, I'll revert out all the `jq` stuff and add `|| true` instead.

vyasr:
@ajschmidt8 thoughts? Just noticed that this PR has been sitting for a bit.

ajschmidt8 (approved):

tools/rapids-configure-conda-channels
```diff
@@ -1,11 +1,29 @@
 #!/bin/bash
 # A utility script that configures conda channels
 
+conda_channel_in_config() {
+  channel_id=${1:?err}
+  conda config --json --get channels \
+  | jq -r --arg c "${channel_id}" '."get"."channels" | any(. == $c )'
```
ajschmidt8, line 7: `jq` has a `--exit-status`/`-e` flag for converting booleans to exit codes.

if you use that flag here, then you can update the calling lines to just look like this:

```sh
if conda_channel_in_config "${channel_id}"; then
# ...
```

also, `jq` doesn't require double quotations when you're accessing keys.

e.g. `'."get"."channels"'` can be just `'.get.channels'`, which helps with readability.

finally, you can avoid having to pass `--arg c` by exporting `channel_id` (e.g. `export channel_id=${1:?err}`) and using `env.channel_id` in `jq` like this:

```sh
jq '.get.channels | any(. == env.channel_id)'
```

ajschmidt8 (approved):
approved with some optional comments/suggestions

ajschmidt8:
also, i noticed this script is missing our standard `set -euo pipefail`. not sure if it's worth adding in this PR.

tools/rapids-configure-conda-channels
jameslamb: Thanks, did not know about `env.{whatever}`!

I just pushed some changes implementing most of this. The one difference is I'm proposing this:

```shell
channel_id="${channel_id}" jq ...
```

instead of

```shell
export channel_id="${channeld_id}"
jq ...
```

So that function doesn't have side-effects observable outside of it (e.g. setting `channel_id` in global scope).

jameslamb:
> 
also, i noticed this script is missing our standard set -euo pipefail. not sure if it's worth adding in this PR.

I just added this in https://github.com/rapidsai/gha-tools/pull/104/commits/f1464ce274d5cca1b406adbea3b221101b5f73d4

Tested my new changes following the "How I tested this" steps in the description of the PR, all seemed to still work the expected way.

I'll wait til you have another chance to review before I merge this @ajschmidt8

ajschmidt8 (approved):

jameslamb:
Thanks again for the thorough review and for teaching me some new `jq` things @ajschmidt8 !

PR #105, opened 2024-04-30
support cpp wheels in 'rapids-wheels-anaconda'

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/44.

Proposes removing hardcoding of `_python_` in the search for artifact names performed by `rapids-wheels-anaconda`, to enable uploading of C++ wheels (which will have `_cpp_` in artifact names).

## Notes for Reviewers

I think this would be safe to merge right now... it should be totally backwards-compatible because it defaults to `_python_` when a `pkg_type` isn't supplied.

To see how the entire S3 object key is computed, also see this:

https://github.com/rapidsai/gha-tools/blob/4c878b81203e29eea312b5774868b83c1b931466/tools/rapids-s3-path#L8-L15

Related to:

* https://github.com/rapidsai/build-planning/issues/57
* https://github.com/rapidsai/ucxx/pull/226
* https://github.com/rapidsai/ucx-wheels/pull/1

### DIFF
--- a/tools/rapids-wheels-anaconda
+++ b/tools/rapids-wheels-anaconda
@@ -3,6 +3,13 @@
 
 # Positional Arguments:
 #   1) wheel name
+#   2) package type (one of: 'cpp', 'python')
+#
+# [usage]
+#
+#   # upload any wheels found in CI artifacts with names like '*wheel_python_sparkly-unicorn*.tar.gz'
+#   rapids-wheels-anaconda 'sparkly-unicorn' 'python'
+#
 
 set -exou pipefail
 source rapids-constants
@@ -13,8 +20,9 @@ if [ -z "$1" ]; then
   exit 1
 fi
 WHEEL_NAME="$1"
+PKG_TYPE="${2:-python}"
 
-WHEEL_SEARCH_KEY="wheel_python_${WHEEL_NAME}"
+WHEEL_SEARCH_KEY="wheel_${PKG_TYPE}_${WHEEL_NAME}"
 
 WHEEL_DIR="./dist"
 mkdir -p "${WHEEL_DIR}"
--- a/tools/rapids-wheels-anaconda
+++ b/tools/rapids-wheels-anaconda
@@ -22,6 +22,17 @@ fi
 WHEEL_NAME="$1"
 PKG_TYPE="${2:-python}"
 
+case "${PKG_TYPE}" in
+  cpp)
+    ;;
+  python)
+    ;;
+  *)
+    rapids-echo-stderr 'Pass one of the following package types: "cpp", "python"'
+    exit 1
+    ;;
+esac
+
 WHEEL_SEARCH_KEY="wheel_${PKG_TYPE}_${WHEEL_NAME}"
 
 WHEEL_DIR="./dist"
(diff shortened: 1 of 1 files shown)


vyasr:
Testing in https://github.com/rapidsai/ucx-wheels/pull/2

tools/rapids-wheels-anaconda
```diff
@@ -3,6 +3,13 @@
 
 # Positional Arguments:
 #   1) wheel name
+#   2) package type (one of: 'cpp', 'python')
```
msarahan, line 6:
```suggestion
#   2) package type (one of: 'cpp', 'python'). If not provided, defaults to 'python' for compatibility with older code where python was the only behavior.
```

msarahan (approved):
approved pending results from Vyas' testing

vyasr (approved):
LGTM! We have a test confirming that this works.

ajschmidt8 (approved):
LGTM.

Do you think it's worth adding a `case` statement to ensure `PKG_TYPE` is `cpp` or `python`?

Similar to what we do here:

https://github.com/rapidsai/gha-tools/blob/4c878b81203e29eea312b5774868b83c1b931466/tools/rapids-download-conda-from-s3#L11-L20

jameslamb:
Great suggestion, thanks @ajschmidt8. Just pushed https://github.com/rapidsai/gha-tools/pull/105/commits/ad0e188510bc0b9be9bea5b3b9b0e67da85a60ac adding that.

Tested it by adding one of these after the `case` block

```shell
echo "--- ${PKG_TYPE}"
exit 6
```

and then a few different combinations

```shell
# fails with this error
PATH="${PWD}/tools:${PATH}" rapids-wheels-anaconda beepboop cpp2

# succeeds
PATH="${PWD}/tools:${PATH}" rapids-wheels-anaconda beepboop cpp

# succeeds
PATH="${PWD}/tools:${PATH}" rapids-wheels-anaconda beepboop python

# succeeds, defaults to 'python'
PATH="${PWD}/tools:${PATH}" rapids-wheels-anaconda beepboop
```

tools/rapids-wheels-anaconda
```diff
@@ -3,6 +3,13 @@
...
 #   1) wheel name
+#   2) package type (one of: 'cpp', 'python'). If not provided, defaults to 'python' for compatibility with older code where python was the only behavior.
+#
+# [usage]
+#
+#   # upload any wheels found in CI artifacts with names like '*wheel_python_sparkly-unicorn*.tar.gz'
+#   rapids-wheels-anaconda 'sparkly-unicorn' 'python'
+#
 
 set -exou pipefail
```
ajschmidt8, line 14:
```suggestion
set -eou pipefail
```

last request.

I usually try to omit the `-x` since it makes the logs very noisy.

I know you didn't add this flag in this PR, but I think it's still worth removing.

jameslamb: Sure, fine with me! For a script this size, I'm indifferent about it 🤷🏻

ajschmidt8 (approved):
merge away!

jameslamb:
Thanks for the reviews!

PR #109, opened 2024-06-17
Make `rapids-print-env` compatible with non-`conda` images

ajschmidt8:
This PR ensures that the `rapids-print-env` command is compatible with non-`conda` CI images.

This ensures the command can be run in `ci-wheel` images.

### DIFF
--- a/tools/rapids-print-env
+++ b/tools/rapids-print-env
@@ -13,7 +13,7 @@ arch
 rapids-logger "Check python version"
 python --version
 
-if conda &> /dev/null; then
+if command -v conda > /dev/null; then
   rapids-logger "Check conda environment"
   conda info
   conda config --show-sources
--- a/tools/rapids-print-env
+++ b/tools/rapids-print-env
@@ -13,7 +13,9 @@ arch
 rapids-logger "Check python version"
 python --version
 
-rapids-logger "Check conda environment"
-conda info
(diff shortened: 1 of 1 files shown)


tools/rapids-print-env
```diff
@@ -13,7 +13,9 @@ arch
 rapids-logger "Check python version"
 python --version
 
-rapids-logger "Check conda environment"
-conda info
-conda config --show-sources
-conda list --show-channel-urls
+if conda &> /dev/null; then
```
bdice, line 16: I think we want to use something like this?
```suggestion
if command -v conda &> /dev/null; then
```

ajschmidt8: Is there any advantage to using `command -v` vs. just trying the `conda` command?

bdice: We should not actually invoke `conda`.

`command -v` is cheaper (basically zero time vs. 0.5 seconds to invoke `conda`).

ajschmidt8: I see. Okay. Good point. I will test the change locally real quick

ajschmidt8: updated in [988969c](https://github.com/rapidsai/gha-tools/pull/109/commits/988969cfa20643bfbfba6ed19c15c68938c63eb8)

PR #110, opened 2024-06-28
update is-release-build script to accommodate varying tag formats

AyodeAwe:
A number of non-RAPIDS repositories which run releases trigger their releases via tags, much like RAPIDS, except that their tag formats are sometimes different from RAPIDS. This PR updates the `is-release-build` tool to accommodate these differences.


AyodeAwe:
Closing in favor of https://github.com/rapidsai/gha-tools/pull/111

PR #111, opened 2024-06-28
consider all tag refs builds as release builds

AyodeAwe:
This PR updates the `rapids-is-release-build` tool to mark all builds triggered from tag refs as release builds.

All RAPIDS libraries releases are currently triggered via tags. Other repositories such as `pynvjitlink`, `rapids-build-backend`, `rapids-metadata`, `ucxx`, `ucx-py`, `jupyterlab-nvdashboard` have only their releases triggered via tags as well so this shouldn't break anything.

cc; @raydouglass

### DIFF
--- a/tools/rapids-is-release-build
+++ b/tools/rapids-is-release-build
@@ -8,7 +8,7 @@
 set -e
 export RAPIDS_SCRIPT_NAME="rapids-is-release-build"
 
-if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
+if [[ "${GITHUB_REF}" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
   rapids-echo-stderr "is release build"
   exit 0
 fi
--- a/tools/rapids-is-release-build
+++ b/tools/rapids-is-release-build
@@ -8,9 +8,10 @@
 set -e
 export RAPIDS_SCRIPT_NAME="rapids-is-release-build"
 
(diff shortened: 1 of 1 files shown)


tools/rapids-is-release-build
```diff
@@ -8,9 +8,10 @@
 set -e
 export RAPIDS_SCRIPT_NAME="rapids-is-release-build"
 
-if [[ "${GITHUB_REF}" =~ ^refs/tags/v[0-9]{1,2}.[0-9]{2}.[0-9]{2}$ ]]; then
+if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
```
raydouglass, line 11: Need to exclude alpha tags (like https://github.com/rapidsai/cudf/releases/tag/v24.08.00a).

Maybe keep the regex but allow for any number of digits?
```suggestion
if [[ "${GITHUB_REF}" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```

AyodeAwe: @raydouglass this should work no?

`if [[ "${GITHUB_REF}" == refs/tags/v* && "${GITHUB_REF}" != *a ]]; then`

AyodeAwe: Hmm, thinking about it, that sure would not work for tags like `v24.08.00b` or `v24.08.00c` etc

raydouglass: Yeah, a `b` suffix was used in the distant past (https://github.com/rapidsai/cudf/tree/v0.12.0b) and it's unlikely we would use it again, but I think this match should be a strict as possible. 

We can add exceptions if needed later.

raydouglass (approved):

PR #118, opened 2024-09-25
Add otel wrapper

msarahan:
This is designed to lay the groundwork for enabling all build jobs to enable telemetry when they are ready. It works by wrapping commands in a way that runs the command with opentelemetry-instrument, which enables the no-code monkeypatching stuff to take effect. 

https://github.com/rapidsai/gha-tools/compare/main...msarahan:gha-tools:add-otel-wrapper?expand=1#diff-f1f054b2906bfd36ad706ed2fa6aa028fa529e65b25167e4ca7ca45546d59ed8R33

The functionality of this PR is ideally non-invasive, because it only becomes active when a particular executable, otel-cli, is available.

Jobs that wish to capture telemetry should:
* Set `RAPIDS_SERVICE_NAME` to the desired descriptor of the job running this command (perhaps the job name including build matrix?). This is used as the "service name" in telemetry records.
* Optionally set `RAPIDS_OTEL_ATTRS` to a comma-separated key-value string. These attrs are filterable and are intended for use in later aggregating data. The build scripts should set their own attrs, but if there are additional identifying attributes you want to filter on, you can set them here.

These really only apply to the conda builds, because wheel build scripts are more per-repo. Hopefully the wrapper script can serve in those areas.


msarahan:
superseded by #119

PR #120, opened 2024-10-14
add opentelemery helper scripts (#119)

msarahan:


tools/rapids-get-telemetry-traceparent
```diff
@@ -11,15 +11,22 @@
 # A step name MAY be provided as the second argument. If it is specified, the output corresponds to
 #     the step within the context of its job.
 
-JOB_NAME=$1
 STEP_NAME=${2:-}
```
msarahan, line 14: Should be $1

tools/rapids-conda-retry
```diff
@@ -67,7 +67,9 @@ condaCmd=${RAPIDS_CONDA_EXE:=conda}
 #    needToRetry: 1 if the command should be retried, 0 if it should not be
 function runConda {
     # shellcheck disable=SC2086
-    ${condaCmd} ${args} 2>&1| tee "${outfile}"
+    # RAPIDS_OTEL_WRAPPER is optionally passed in as an env var. It is
+    # used to instrument conda-build or mambabuild for finer-grained spans.
+    ${RAPIDS_OTEL_WRAPPER:-} ${condaCmd} ${args} 2>&1| tee "${outfile}"
```
msarahan, line 72: This makes the wrapper opt-in, which I hope avoids any lingering possible issues that occurred last time we tried to add this. The wrapper won't be present at all unless it is added as an environment variable. This will happen in shared-workflows files.

ajschmidt8 (approved):

PR #125, opened 2024-11-22
Check if in CI before running rapids-package-name

vyasr:
rapids-package-name expects the various variables set in CI to exist to work (and there's no reason to prompt for those locally like we do when downloading since we'll never upload anyway).


jameslamb (approved):
Ok yeah, makes sense to me.

PR #128, opened 2024-12-17
Allow test package uploads

raydouglass:
Allow for test conda packages to be uploaded to make it easier for users to run tests or benchmarks.


bdice (approved):

tools/rapids-find-anaconda-uploads.py
```diff
@@ -54,8 +54,6 @@ def file_filter_fn(file_path):
     filename = path.basename(file_path)
     pkg_name = get_pkg_name_from_filename(filename)
 
-    if is_test_pkg(pkg_name):
```
bdice, line 57: Should we also remove the function `is_test_pkg`?

PR #131, opened 2025-01-03
Add ability to pull a wheel artifact as well as a conda one

vyasr:
This generalizes the preexisting `rapids-get-pr-conda-artifact` to support wheels via `rapids-get-pr-artifact`. The former is now a trivial passthrough to the latter for backwards compatibility.


jameslamb:
Have you seen that there's already a `rapidsa-get-pr-wheel-artifact` that does this? Would that meet the need for whatever you're trying to do?

https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-get-pr-wheel-artifact

I've found it helpful for testing PRs that rely on wheels from other PRs / repos.

vyasr:
Wow. Well... no :cry: my local copy of gha-tools must have been very out of date, and I haven't done nearly as much wheel development of late. Ah well. I guess the question now is whether it's worth using this PR as an opportunity for cleanup to make `rapids-get-pr-conda-artifact` and `rapids-get-pr-wheel-artifact` both call the new function. The two scripts are basically copies of each other and now that I've already made this PR it seems like we might as well do to the work to unify them. WDYT? If you'd rather skip that idea I can close this PR.

jameslamb:
I'm supportive of consolidating more of the shared logic into one place like this, if you want to spend the time on it! Happy to review and help test.

If `rapids-get-pr-artifact` is intended to be just for code-sharing and not called explicitly, I think for clarity it should be prefixed with an underscore, like `_rapids-download-from-s3`: https://github.com/rapidsai/gha-tools/blob/main/tools/_rapids-download-from-s3

vyasr:
Done.

tools/_rapids-get-pr-artifact
```diff
@@ -0,0 +1,32 @@
...
+#   5) [optional] commit hash, to get the artifact for a specific commit
+#
+# If fetching a wheel, RAPIDS_PY_WHEEL_NAME must be set to indicate the package
+# name to fetch.
+# Example Usage:
+#   _rapids-get-pr-artifact rmm 1095 cpp
+set -euo pipefail
+
+repo="$1"
+pr="$2"
```
jameslamb, line 18:
```suggestion
pr="$2"
package_type="${3}"
package_format="${4}"
commit="${5:-}"
```

`pkg_type` / `package_type` is used in other scripts here to mean "`cpp` or `python`", so I think it's a little confusing for it to have a meaning in this script that's like "`cpp` or `python` AND `wheel` or `conda`".

I also think mapping all of the positional args to informatively-named variables at the top makes the rest of the code a bit easier to read. I know some of the code here was copied from an implementation already not doing that, but I noticed this because `_rapids-get-pr-artifact` is now re-using `${4}` in multiple places.

Would you consider the following?

1. doing all the mapping of positional args to named variables at the top
2. using `package_format` for the "`wheel` or `conda`" variable

jameslamb (approved):
Just one minor suggestion, do what you want with it, otherwise this looks good to me!

tools/_rapids-get-pr-artifact
vyasr: Yeah that all seems fine to me.

vyasr: Done.

PR #132, opened 2025-01-28
Add `rapids-pip-retry` in the same vein as `rapids-conda-retry`

gforsyth:
A naive translation of `rapids-conda-retry` that currently only flags for `ERROR: THESE PACKAGES DO NOT MATCH THE HASHES` messages that can break install attempts if the network connection is interrupted mid-download.

### DIFF
--- a/tools/rapids-pip-retry
+++ b/tools/rapids-pip-retry
@@ -27,7 +27,7 @@ pipretry_help="
 rapids-pip-retry options:
 
    --pipretry_max_retries=n        Retry the pip command at most n times (default is 3)
-   --pipretry_sleep_interval=n     Sleep n seconds between retries (default is 5)
+   --pipretry_sleep_interval=n     Sleep n seconds between retries (default is 10)
 
 ALSO rapids-pip-retry options can be set using the following env vars:
 
--- a/tools/rapids-pip-retry
+++ b/tools/rapids-pip-retry
@@ -1,0 +1,139 @@
+#!/bin/bash
+#
+# rapids-pip-retry
+#
+# wrapper for pip that retries the command after a hash mismatch error that
+# (usually) results from an interrupted network connection
+#
+# Example usage:
+# $ rapids-pip-retry install nvidia-cublas-cu11
+#
+# Configurable options are set using the following env vars:
+#
+# RAPIDS_PIP_EXE             - override the pip executable
+#                              Default is "pip"
+#
+# RAPIDS_PIP_RETRY_MAX       - set to a positive integer to set the max number of retry
+#                              attempts (attempts after the initial try).
+#                              Default is 3 retries
+#
+# RAPIDS_PIP_RETRY_SLEEP     - set to a positive integer to set the duration, in
+#                              seconds, to wait between retries.
+#                              Default is a 10 second sleep
+set -o pipefail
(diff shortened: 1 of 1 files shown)


bdice (approved):
Looks fine to me! Let's focus on deploying this in raft/cuvs/cuml/cugraph first, because they use the CUDA math library wheels.

tools/rapids-pip-retry
```diff
@@ -0,0 +1,139 @@
...
+#                              seconds, to wait between retries.
+#                              Default is a 10 second sleep
+set -o pipefail
+export RAPIDS_SCRIPT_NAME="rapids-pip-retry"
+
+pipretry_help="
+rapids-pip-retry options:
+
+   --pipretry_max_retries=n        Retry the pip command at most n times (default is 3)
+   --pipretry_sleep_interval=n     Sleep n seconds between retries (default is 5)
```
robertmaynard, line 30:
```suggestion
   --pipretry_sleep_interval=n     Sleep n seconds between retries (default is 10)
```

tools/rapids-pip-retry
```diff
@@ -0,0 +1,139 @@
...
+# shellcheck disable=SC2086
+runPip ${args}
+
+# Retry loop, only if needed
+while (( needToRetry == 1 )) && \
+      (( retries < max_retries )); do
+
+   retries=$(( retries + 1 ))
+   rapids-echo-stderr "Waiting, retry ${retries} of ${max_retries} -> sleeping for ${sleep_interval} seconds..."
+   sleep "${sleep_interval}"
```
robertmaynard, line 131: We should add some random jitter to the sleep interval. Otherwise we can have all of our workers fail to download, and wait the exact amount of time and re-hammer the download.

robertmaynard (commented):
Primary concern is having conccurent workers with the same sleep interval. I think having a MIN and MAX sleep duration and selecting a random number between them is better.

vyasr:
@gforsyth could we test this PR by opening a PR to one of our repos (like cuvs) and switching it over to use this? That's usually the best way to validate these tools (and also makes it easy to get repos updated once these PRs merge).

robertmaynard (approved):

gforsyth:
`cuvs` had some unreleated build failures, so I'm also testing in `raft` to see how effective this is

gforsyth:
Those are failures that can be re-attempted via `pip retries` like @jameslamb set up in rapidsai/ci-imgs#233 -- there may be a separate problem there, but this utility will (currently) only help with the hash mismatches that seem to result from a broken network connection happening at an inopportune time, like in https://github.com/rapidsai/raft/actions/runs/13005031794/job/36270469163#step:10:118

That's the error I haven't seen re-occur (and that I was trying to trigger in the `raft` and `cuvs` PRs without success)

vyasr:
I saw that the test PRs were closed. I guess they validated that this tool doesn't break anything but not that it's fixing anything either. I suggest that if we're not seeing anything we try merging and let one round of nightlies go since that's where we've had the most issues. Then we'll have a bit more real data.

gforsyth:
I think it's worth trying.  At the very least, we've confirmed it doesn't interfere with any `pip install` commands and it might help.

PR #133, opened 2025-01-31
Retry 'Unexpected error [0-9]* on netlink descriptor [0-9]*'

bdice:
Recently we have had several nightly failures due to network issues. These errors appeared as:

https://github.com/rapidsai/ucxx/actions/runs/13067745825/job/36462909749#step:8:3868
```
Unexpected error 9 on netlink descriptor 7.
```

https://github.com/rapidsai/rmm/actions/runs/13066928144/job/36460789434#step:8:1600
```
Unexpected error 9 on netlink descriptor 9.
```

This has come up before, as well. See the following links:
- https://github.com/rapidsai/ci-imgs/issues/228#issuecomment-2597141388
- https://github.com/rapidsai/ci-imgs/issues/162#issuecomment-2596845806
- https://github.com/rapidsai/cudf/pull/17585#pullrequestreview-2512061190
- https://github.com/rapidsai/ci-imgs/pull/219#issuecomment-2596481942

@jameslamb has [pointed to a potential root cause](https://github.com/rapidsai/ci-imgs/pull/219#issuecomment-2596490603):

> I found a couple discussions on the internet suggesting that that particular error can come from non-thread-safe handling of file descriptors:
> - https://github.com/curl/curl/issues/3386#issuecomment-455840319
> - https://stackoverflow.com/a/59615786/3986677

At this point it seems like retrying is the best solution we have available.

### DIFF
--- a/tools/rapids-conda-retry
+++ b/tools/rapids-conda-retry
@@ -119,7 +119,7 @@ function runConda {
         elif grep -q "Timeout was reached" "${outfile}"; then
             retryingMsg="Retrying, found 'Timeout was reached' in output..."
             needToRetry=1
-        elif grep -q "Unexpected error [0-9]* on netlink descriptor [0-9]*" "${outfile}"; then
+        elif grep -q "Unexpected error [0-9]+ on netlink descriptor [0-9]+" "${outfile}"; then
             retryingMsg="Retrying, found 'Unexpected error [0-9]* on netlink descriptor [0-9]*' in output..."
             needToRetry=1
         elif [[ $exitcode -eq 139 ]]; then
--- a/tools/rapids-conda-retry
+++ b/tools/rapids-conda-retry
@@ -119,6 +119,9 @@ function runConda {
         elif grep -q "Timeout was reached" "${outfile}"; then
             retryingMsg="Retrying, found 'Timeout was reached' in output..."
             needToRetry=1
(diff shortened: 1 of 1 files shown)


tools/rapids-conda-retry
```diff
@@ -119,6 +119,9 @@ function runConda {
         elif grep -q "Timeout was reached" "${outfile}"; then
             retryingMsg="Retrying, found 'Timeout was reached' in output..."
             needToRetry=1
+        elif grep -q "Unexpected error [0-9]* on netlink descriptor [0-9]*" "${outfile}"; then
```
msarahan, line 122: Feel free to reject this, as I haven't seen lots of these errors myself. If you expect that there should always be a number, then + is better than *.

```suggestion
        elif grep -q "Unexpected error [0-9]+ on netlink descriptor [0-9]+" "${outfile}"; then
```

msarahan (approved):

tools/rapids-conda-retry
jameslamb: I have always seen a number there, I think the `+` is a good suggestion.

jameslamb (approved):
Thanks for doing this, I agree! This is a retryable error.

I spent some time yesterday trying to even just get this to reliably reproduce and couldn't, but we do keep seeing it in CI.

PR #137, opened 2025-02-03
Adds sourceable scriptlet for setting rattler-build env vars

gforsyth:
`rattler-build` doesn't have builtin support for `GIT_DESCRIBE_NUMBER` and `GIT_DESCRIBE_HASH` so this script sets them so they can be imported via environment variable.


tools/rapids-configure-rattler
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# A utility script that sets variables needed for `rattler-build`
+
+# We only want the last two (of three) entries from splitting the `git describe` output on `-`
+# so we purposefully overwrite `GIT_DESCRIBE_NUMBER` to avoid setting an unwanted variable
```
jameslamb, line 5: Thanks for this comment, because the duplication is confusing! And I do get that anything else we might do has a non-0 chance of conflicting with something else in the environment, e.g. like this:

```shell
IFS=- read -r __ignore_me GIT_DESCRIBE_NUMBER GIT_DESCRIBE_HASH <<< "$(git describe --tags)"
unset __ignore_me
```

I'm not familiar enough with `bash` to know if there's something else that could be done to here, so I think this approach + the comment is a good one.

jameslamb (approved):
Makes sense to me!

I tested this locally in my clone of `ucx-py`, and it did exactly what I'd expected it to:

```shell
git fetch upstream --tags
git describe --tags
# v0.43.00a-7-g58ceab5

IFS=- read -r GIT_DESCRIBE_NUMBER GIT_DESCRIBE_NUMBER GIT_DESCRIBE_HASH <<< "$(git describe --tags)"

echo ${GIT_DESCRIBE_NUMBER}
# 7

echo ${GIT_DESCRIBE_HASH}
# g58ceab5
```

PR #139, opened 2025-02-04
remove rapids-otel-wrap for simplicity

msarahan:
This proved fragile, and the complexity of having this here is not worth the potential future application. It was fragile in the sense that instrumenting conda-build led to bizarre failures that would be deep rabbit holes to investigate.


jameslamb (approved):
thanks for coming back to rip this out!

PR #140, opened 2025-02-04
WIP: add a timeout on rapids-conda-retry and rapids-mamba-retry

### LINKED ISSUE #129: Consider adding a timeout for `mamba install`
This isn't a case I've seen frequently, in fact it is something I've seen for the first time in [this run](https://github.com/rapidsai/ucx-py/actions/runs/12387890462/job/34578058418). For some reason, that one particular `mamba install` never completed (but rerunning worked), and due to that a GPU runner was there idle for 6h until hitting GHA's timeout.

I suggest adding a "safe" timeout to install commands such as `rapids-mamba-retry`, obviously it cannot be too short, but perhaps a somewhat high timeout such as 30m or 60m would prevent this from happening again while still being broadly safe for all our repos.

jameslamb:
Fixes #129 

As described in #129, we have sometimes observed `conda install`, `conda env update`, or similar run indefinitely. This is problematic because it means that a doomed-to-fail job can end up occupying a GPU-enabled CI runner for up to 6 hours ([the hard limit on runtime for a GitHub Actions job](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits)).

Proposes the following 

## Notes for Reviewers

**IF YOU SEE THIS NOTE, THIS IS NOT READY FOR REVIEW**

### How I tested this

TBD

### DIFF
--- a/tools/rapids-conda-retry
+++ b/tools/rapids-conda-retry
@@ -2,12 +2,8 @@
 #
 # rapids-conda-retry
 #
-# wrapper for conda that retries the command after a CondaHTTPError,
-# ChecksumMismatchError, or JSONDecodeError (ideally, any conda error that
-# is normally resolved by retrying)
-#
-# This must be set in order for the script to recognize failing exit codes when
-# output is piped to tee
+# wrapper for conda that retries the command after retryable errors like
+# CondaHTTPError, ChecksumMismatchError, and JSONDecodeError
 #
 # Example usage:
 # $ rapids-conda-retry install cudatoolkit=11.0 rapids=0.16
@@ -23,9 +19,18 @@
 #
 # RAPIDS_CONDA_RETRY_SLEEP     - set to a positive integer to set the duration, in
 #                                seconds, to wait between retries.
-#                                Default is a 10 second sleep
+#                                Default is a 10 second sleep.
 #
+# RAPIDS_CONDA_RETRY_TIMEOUT   - Timeout for each individual retry.
+#                                Positive integers are interpreted as seconds,
+#                                but unit strings like '2h' for "two hours" will also work.
+#                                Default varies based on the command being run.
+#
+
+# This must be set in order for the script to recognize failing exit codes when
+# output is piped to tee
 set -o pipefail
+
 export RAPIDS_SCRIPT_NAME="rapids-conda-retry"
 
 condaretry_help="
@@ -43,6 +48,11 @@ ALSO rapids-conda-retry options can be set using the following env vars:
     RAPIDS_CONDA_RETRY_SLEEP     - set to a positive integer to set the duration, in
                                    seconds, to wait between retries.
                                    Default is a 10 second sleep
+
+    RAPIDS_CONDA_RETRY_TIMEOUT   - Timeout for each individual retry.
+                                   Positive integers are interpreted as seconds,
+                                   but unit strings like '2h' for 'two hours' will also work.
+                                   Default varies based on the command being run.
 ==========
 "
 max_retries=${RAPIDS_CONDA_RETRY_MAX:=3}
@@ -67,7 +77,7 @@ condaCmd=${RAPIDS_CONDA_EXE:=conda}
 #    needToRetry: 1 if the command should be retried, 0 if it should not be
 function runConda {
     # shellcheck disable=SC2086
-    ${condaCmd} ${args} 2>&1| tee "${outfile}"
+    timeout --verbose "${timeout_duration}" ${condaCmd} ${args} 2>&1| tee "${outfile}"
     exitcode=$?
     needToRetry=0
     needToClean=0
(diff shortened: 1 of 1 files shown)


tools/rapids-conda-retry
```diff
@@ -23,9 +19,18 @@
...
+#                                Default is a 10 second sleep.
+#
+# RAPIDS_CONDA_RETRY_TIMEOUT   - Timeout for each individual retry.
+#                                Positive integers are interpreted as seconds,
+#                                but unit strings like '2h' for "two hours" will also work.
+#                                Default varies based on the command being run.
 #
+
+# This must be set in order for the script to recognize failing exit codes when
+# output is piped to tee
```
jameslamb, line 31: I think it was a mistake that this got mixed in with the multi-line comment about arguments... seems like it was supposed to specific to the `set -o pipefail` line.

tools/rapids-conda-retry
```diff
@@ -151,16 +168,17 @@ function runConda {
...
-           (( retries >= max_retries )); then
-            # Catch instance where we run out of retries
-            rapids-echo-stderr "Exiting, reached max retries..."
-        else
-            # Give reason for retry
-            rapids-echo-stderr "${retryingMsg}"
-            if (( needToClean == 1 )); then
-                rapids-echo-stderr "Cleaning tarball cache before retrying..."
-                ${condaCmd} clean --tarballs -y
+        if (( needToRetry == 1 )); then
```
jameslamb, line 171: Re-formatting this because in the previous example, `rapids-echo-stderr "${retryingMsg}"` always ran, resulting in 1 empty log line in the case where `needToRetry` is not `1`.

tools/rapids-conda-retry
```diff
@@ -184,6 +202,24 @@ for arg in "$@"; do
    fi
 done
 
+# Set a default timeout based on command being run.
+#
+# This prevents occupying a CI runner for too long in cases where
+# other timeout mechanisms in 'conda' / 'mamba' are not sufficient to
+# interrupt to a long-running operation.
```
jameslamb, line 209:
```suggestion
# interrupt a long-running operation.
```

tools/rapids-mamba-retry
```diff
@@ -17,7 +17,12 @@
 #                                seconds, to wait between retries.
 #                                Default is a 10 second sleep
 #
-# These are copied to RAPIDS_CONDA_RETRY_MAX and RAPIDS_CONDA_RETRY_SLEEP
+# RAPIDS_MAMBA_RETRY_TIMEOUT   - Timeout for each individual retry.
+#                                Positive integers are interpreted as seconds,
+#                                but unit strings like '2h' for "two hours" will also work.
+#                                Default varies based on the command being run.
+#
+# These are copied to RAPIDS_CONDA_RETRY_MAX, RAPIDS_CONDA_RETRY_SLEEP, and RAPIDS_MAMBA_RETRY_TIMEOUT.
```
jameslamb, line 25:
```suggestion
# These are copied to RAPIDS_CONDA_RETRY_MAX, RAPIDS_CONDA_RETRY_SLEEP, and RAPIDS_CONDA_RETRY_TIMEOUT.
```

tools/rapids-conda-retry
```diff
@@ -130,6 +140,13 @@ function runConda {
             retryingMsg="Retrying, command resulted in a segfault. This may be an intermittent failure..."
             needToRetry=1
             needToClean=1
+        elif [[ $exitcode -eq 124 ]]; then
+            # 'timeout' returns exit code 124 when the timeout is exceeded.
+            # ref: https://man7.org/linux/man-pages/man1/timeout.1.html
+            exitMsg="Exiting, command exited with status 124 which often indicates a timeout (configured timeout='${timeout_duration}')."
+            exitMsg+=" To increase this timeout, set env variable RAPIDS_CONDA_RETRY_TIMEOUT."
+            rapids-echo-stderr "${exitMsg}"
+            needToRetry=0
```
jameslamb, line 149: Calling your attention to an important design decision.

**proposal: timeouts should not be retried**

Making a timeout non-retryable makes the timeout setting easier to understand... it means there's no meaningful difference between these:

- *"fail if `conda install` runs for more than 45 minutes"*
- *"fail if `rapids-conda-retry install` runs for more than 45 minutes".*

And note that by default, all the timeouts around things like SSL verification, network connections, etc. will still end up getting retried, because those raise errors like `CondaHTTPError` or `Timeout was reached`, with default configurations on the order of seconds and not 10s of minutes.

This blunter mechanism covers other issues that might generally not be retryable, like solver deadlock or waiting indefinitely for the filesystem to respond.

If we were to apply a timeout to **the entire `rapids-{conda,mamba}-retry`** operation instead, it'd be more complicated to get this right. Because you'd have to set a value that covers the total runtime of all retries.

CONTRIBUTING.md
```diff
@@ -65,3 +65,33 @@ Since the current working directory is a standard RAPIDS repository, the CI scri
 ```sh
 ./ci/test_python.sh
 ```
+
+## Testing in CI
+
+The tools here are all executable, so testing an alternative branch just requirements downloading
```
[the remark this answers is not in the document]
jameslamb: 🤦🏻 thank you for catching this.

msarahan (approved):

jameslamb:
Seeing no other comments here, I'm going to merge this. I'm confident in it, based on the details in the "How I tested this" section of the description.

bdice (commented):
Thanks @jameslamb, all looks good to me. Apologies, this PR got lost on my queue. Thanks for merging!

PR #142, opened 2025-02-11
Remove `rapids-configure-rattler`

gforsyth:
I added this to support `rattler-build` but we are using the `rattler`
built-in `git` functions instead so this is no longer needed.


jameslamb (approved):
Thanks for coming back and cleaning it up!

PR #143, opened 2025-02-11
Add `rapids-rattler-channel-string` tool

gforsyth:
This is designed to be `source`d so that the channel list can be
`echo`ed to stdout while saving the properly formatted channel string in
an exported variable.

xref rapidsai/rmm#1796

### DIFF
--- a/tools/rapids-rattler-channel-string
+++ b/tools/rapids-rattler-channel-string
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+RAPIDS_CHANNEL="rapidsai-nightly"
+DASK_CHANNEL="dask/label/dev"
+
+# Replace dev/nightly channels if build is a release build
+if rapids-is-release-build; then
+  RAPIDS_CHANNEL="rapidsai"
+  DASK_CHANNEL=""
+fi
+
+channels=("$RAPIDS_CHANNEL" "$DASK_CHANNEL" "conda-forge" "nvidia")
+
+_add_c_prefix() {
+  for channel in "${channels[@]}"; do
+    # Only echo out a channel if it is non-empty
+    if [[ $channel ]]; then
+      echo -n "-c $channel "
+    fi
+  done
+}
+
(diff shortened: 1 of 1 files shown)


msarahan (approved):

tools/rapids-rattler-channel-string
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+RAPIDS_CHANNEL="rapidsai-nightly"
```
msarahan, line 3:
```suggestion
RAPIDS_CHANNEL="${RAPIDS_CHANNEL:-rapidsai-nightly}"
```
Minor thing, but might make it easier to toy with alternate channels.

bdice: I’m guessing that anywhere we would use this, we would have (and would want) the ability to override the whole channel set. I’m not sure if this flexibility buys us very much - but happy to hear others’ thoughts as well.

bdice (approved):

tools/rapids-rattler-channel-string
msarahan: I do mean other env vars, too, but I agree that I'm not sure that the flexibility buys us much. Still, the NVAIE LWS builds have made me extremely grateful for this kind of flexibility in various places. Even though the normal open-source-facing builds would never use this stuff, it may be useful in other scenarios.

gforsyth: In this _particular_ context, I think the escape valve is setting `$RATTLER_CHANNELS` explicitly, but I'm happy to revisit if we find ourselves needing to tweak it!

gforsyth:
Ugh, I messed this up.  We want to `export` an array so we can expand it without implicit word splitting.  Fix incoming

tools/rapids-rattler-channel-string
vyasr: I agree, I think we'll always be in a position to just not use this tool and/or replace the channels manually.

PR #146, opened 2025-02-21
refactor(rattler): only include `nvidia` channel for CUDA11 builds

gforsyth:
We don't need the `nvidia` channel for CUDA12 and it can complicate solves, so we only include it for CUDA11.

xref rapidsai/build-planning#47

### DIFF
--- a/tools/rapids-rattler-channel-string
+++ b/tools/rapids-rattler-channel-string
@@ -2,7 +2,7 @@
 
 RAPIDS_CHANNEL="rapidsai-nightly"
 DASK_CHANNEL="dask/label/dev"
-NVIDIA_CHANNEL="nvidia"
+NVIDIA_CHANNEL=""
 
 # Replace dev/nightly channels if build is a release build
 if rapids-is-release-build; then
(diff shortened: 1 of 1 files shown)


jameslamb (commented):
Does `rattler-build` also consider the system-level settings for conda channels?

If it does, then this change won't be sufficient, because we put the `nvidia` channel into the `.condarc` in `rapidsai/ci-conda` images:

https://github.com/rapidsai/ci-imgs/blob/028ad7cd8ed3c0a0bcddfe3888938689a2eacd5b/context/condarc.tmpl#L7

gforsyth:
> Does `rattler-build` also consider the system-level settings for conda channels?

Luckily (?) `rattler-build` completely ignores `.condarc` -- I think we'll probably remove the `.condarc` entirely once we have everything ported over

jameslamb:
Ok, gotchu!

> I think we'll probably remove the `.condarc` entirely once we have everything ported over

That file configures more than just channels, and those images are used for more than just package builds (like tests and docs builds), so I'd be careful about that. But anyway, can be talked about in future issues/PRs.

tools/rapids-rattler-channel-string
```diff
@@ -2,20 +2,27 @@
...
   RAPIDS_CHANNEL="rapidsai"
   DASK_CHANNEL=""
 fi
 
-channels=("$RAPIDS_CHANNEL" "$DASK_CHANNEL" "conda-forge" "nvidia")
+# Only include nvidia channel if we're on cuda11
+IFS='.' read -ra CUDA_VERSION_ARRAY <<< "$RAPIDS_CUDA_VERSION"
+if [ "${CUDA_VERSION_ARRAY[0]}" -eq "12" ]; then
+  NVIDIA_CHANNEL=""
+fi
```
jameslamb, line 17: Can you flip this, like this?

```shell
NVIDIA_CHANNEL=""
if [ "${CUDA_VERSION_ARRAY[0]}" -eq "11" ]; then
  NVIDIA_CHANNEL="nvidia"
fi
```

That way, it'd work in a future with newer major CUDA versions than 12.

jameslamb (approved):
Left one suggestion, otherwise this looks ok to me. It would be good to test on a repo (like `cugraph` or `raft`) before we merge this.

tools/rapids-rattler-channel-string
gforsyth: yep, that's a good shout -- flipped

bdice (approved):

PR #149, opened 2025-02-26
Add rapids-telemetry-record

bdice:
This PR proposes a design for `rapids-telemetry-record` so that RAPIDS repositories do not need to hardcode a path for telemetry log artifact output.

### DIFF
--- a/tools/rapids-telemetry-record
+++ b/tools/rapids-telemetry-record
@@ -14,7 +14,7 @@ output_path="${GITHUB_WORKSPACE:-"."}/telemetry-artifacts/${output_file}"
 
 # Run the command, redirecting both stdout and stderr to tee.
 # This writes the combined output to the specified file while also printing it.
-"$@" 2>&1 | tee "$output_file"
+"$@" 2>&1 | tee "${output_path}"
 
 # Exit with the same status as the command that was run.
-exit ${PIPESTATUS[0]}
+exit "${PIPESTATUS[0]}"
--- a/tools/rapids-telemetry-record
+++ b/tools/rapids-telemetry-record
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Check for at least 2 arguments: filename and command
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 filename command [arguments...]"
+    exit 1
+fi
+
+output_file="$1"
+shift
+
+# TODO: Make this a temporary path rather than writing in the GITHUB_WORKSPACE.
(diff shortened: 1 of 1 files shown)


tools/rapids-telemetry-record
```diff
@@ -0,0 +1,20 @@
...
+
+output_file="$1"
+shift
+
+# TODO: Make this a temporary path rather than writing in the GITHUB_WORKSPACE.
+output_path="${GITHUB_WORKSPACE:-"."}/telemetry-artifacts/${output_file}"
+
+# Run the command, redirecting both stdout and stderr to tee.
+# This writes the combined output to the specified file while also printing it.
+"$@" 2>&1 | tee "$output_file"
```
bdice, line 17:
```suggestion
"$@" 2>&1 | tee "${output_path}"
```

tools/rapids-telemetry-record
```diff
@@ -0,0 +1,20 @@
...
+
+# Exit with the same status as the command that was run.
+exit ${PIPESTATUS[0]}
```
bdice, line 20:
```suggestion
exit "${PIPESTATUS[0]}"
```

tools/rapids-telemetry-record
```diff
@@ -0,0 +1,20 @@
...
+# Check for at least 2 arguments: filename and command
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 filename command [arguments...]"
+    exit 1
+fi
+
+output_file="$1"
+shift
+
+# TODO: Make this a temporary path rather than writing in the GITHUB_WORKSPACE.
```
bdice, line 12: @msarahan I'd like your thoughts on this. Could we move telemetry artifacts to use a temporary directory? Thankfully, merging this PR in the current state should isolate those changes from the library repository level -- it would be only a matter of aligning `shared-actions` and `gha-tools`.

bdice: Perhaps we could even make this run `mkdir -p` in a way that is idempotent, and removes the need for `rapids-telemetry-setup`. And/or we could write the telemetry artifact path to `${GITHUB_ENV}` for use in later steps? Not sure.

msarahan: Yes, you can write the output path to GITHUB_ENV. That should work without issue, as these paths are not carried between workflows.

msarahan (approved):

tools/rapids-telemetry-record
bdice: I’ll see if we can revisit this in a follow-up. This is sufficient for now.

PR #152, opened 2025-03-11
Adding scripts to download and upload artifacts to the Github Artifact Store

VenkateshJaya:
Adding new scripts which use the Github Artifact Store to upload and download build artifacts towards https://github.com/rapidsai/ops/issues/2982. No existing scripts or functionality is altered.


jameslamb (commented):
I'm going to join here as a reviewer alongside @gforsyth since you and I have been talking about this work and since I'd done an earlier version of it.

To start... how did you test this? Could we see some PRs showing how these would be used?

VenkateshJaya:
Hey @jameslamb, Here are a couple of PRs on RMM and cuDF which use the new `gha-tools` scripts which are being added now
1. RMM: https://github.com/rapidsai/rmm/pull/1810
2. cuDF: https://github.com/rapidsai/cudf/pull/18090

To test these changes, I used code from my `gha-tools` fork which is overwriting the existing `gha-tools` folder on the Docker image on this branch of the `shared-workflows` repo: https://github.com/rapidsai/shared-workflows/tree/test-github-artifacts

Let me know if you need any additional details, I'll be happy to discuss them with you!

tools/rapids-download-wheels-from-github
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# A utility script that downloads wheel artifacts from Github Artifacts to a temporary directory and returns the path to the directory
+set -eo pipefail
+
+source rapids-prompt-local-repo-config
+
+# For legacy reasons, allow this script to be run without the pkg_type being the first arg
+pkg_name="$(rapids-package-name "wheel_python")"
```
gforsyth, line 8: Do we care about these "legacy reasons" in a new wrapper script?  I think we should eliminate support for older patterns since this is a new pattern.

tools/rapids-get-pr-conda-artifact-github
```diff
@@ -0,0 +1,46 @@
...
+#   rapids-get-pr-conda-artifact-github rmm 1095 cpp
+set -euo pipefail
+
+repo="${1}"
+pr="${2}"
+package_type="${3}"
+commit="${4:-}"
+
+# Create a temporary directory for the artifact
+unzip_dest=$(mktemp -d)
```
gforsyth, line 21: Could we add this as an optional additional argument for dev use to pull the packages down elsewhere?  Or alternatively, because argument munging in bash is hard, we could support setting an env-var to determine download location?

tools/rapids-get-pr-wheel-artifact-github
```diff
@@ -0,0 +1,53 @@
...
+commit="${4:-}"
+
+# Check if RAPIDS_PY_WHEEL_NAME is set when fetching a wheel
+if [[ -z "${RAPIDS_PY_WHEEL_NAME:+placeholder}" ]]; then
+    echo "The RAPIDS_PY_WHEEL_NAME variable must be set when fetching a wheel artifact"
+    exit 1
+fi
+
+# Create a temporary directory for the artifact
+unzip_dest=$(mktemp -d)
```
gforsyth, line 28: Could we add this as an optional additional argument for dev use to pull the packages down elsewhere?  Or alternatively, because argument munging in bash is hard, we could support setting an env-var to determine download location?

tools/rapids-github-run-id
```diff
@@ -0,0 +1,25 @@
...
+    run_id=$(gh run list --repo "${RAPIDS_REPOSITORY}" --branch "${RAPIDS_REF_NAME}" --commit "${RAPIDS_SHA}" --json databaseId --jq '.[0] | .databaseId')
+    ;;
+  nightly)
+    run_id=$(gh run list --repo "${RAPIDS_REPOSITORY}" --branch "${RAPIDS_REF_NAME}" --commit "${RAPIDS_SHA}" --json databaseId --jq '.[0] | .databaseId')
+    ;;
+  *)
+    rapids-echo-stderr "please pass a valid RAPIDS_BUILD_TYPE"
+    exit 1
+    ;;
+esac
```
gforsyth, line 23:
```suggestion
case "${RAPIDS_BUILD_TYPE}" in
  pull-request | branch | nightly)
    run_id=$(gh run list --repo "${RAPIDS_REPOSITORY}" --branch "${RAPIDS_REF_NAME}" --commit "${RAPIDS_SHA}" --json databaseId --jq '.[0] | .databaseId')
    ;;
  *)
    rapids-echo-stderr "please pass a valid RAPIDS_BUILD_TYPE"
    exit 1
    ;;
esac
```

I think we can simplify this case statement to do the one command for any of the valid build types

tools/rapids-github-run-id
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# A utility script that examines environment variables provided
+# by GitHub Actions to print out a Github Workflow Run ID where the expected artifact
+# should be.
+set -euo pipefail
+source rapids-constants
```
gforsyth, line 6: do we need any of the values in here?

tools/rapids-get-pr-wheel-artifact-github
```diff
@@ -0,0 +1,53 @@
...
+#   rapids-get-pr-wheel-artifact-github rmm 1095 cpp
+set -euo pipefail
+
+repo="${1}"
+pr="${2}"
+package_type="${3}"
+commit="${4:-}"
+
+# Check if RAPIDS_PY_WHEEL_NAME is set when fetching a wheel
+if [[ -z "${RAPIDS_PY_WHEEL_NAME:+placeholder}" ]]; then
```
gforsyth, line 22:
```suggestion
if [[ -z "${RAPIDS_PY_WHEEL_NAME}" ]]; then
```

I don't think `+placeholder` is providing anything here, but let me know if I'm missing something.

tools/rapids-download-from-github
```diff
@@ -0,0 +1,21 @@
...
+#   1) package name to generate s3 path for
+set -euo pipefail
+export RAPIDS_SCRIPT_NAME="rapids-download-from-github"
+
+if [ -z "$1" ]; then
+  rapids-echo-stderr "Must specify input arguments: PKG_NAME"
+  exit 1
+fi
+
+github_run_id="$(rapids-github-run-id)"
```
gforsyth, line 14: Is the assumption here that all of the env-vars that `rapids-github-run-id` needs are already set here?  If so, I think there should be a comment to that effect (or checks that they are extant)

tools/rapids-download-conda-from-github
```diff
@@ -0,0 +1,31 @@
...
+  *)
+    rapids-echo-stderr 'Pass "cpp" or "python" as an argument.'
+    exit 1
+    ;;
+esac
+
+source rapids-prompt-local-repo-config
+
+# Prepare empty dir to extract tarball to
+# channel="/tmp/${pkg_type}_channel"
```
gforsyth, line 25: leftover?

tools/rapids-download-conda-from-github
```diff
@@ -0,0 +1,31 @@
...
+case "${pkg_type}" in
+  cpp)
+    ;&
+  python)
+    ;;
+  *)
+    rapids-echo-stderr 'Pass "cpp" or "python" as an argument.'
+    exit 1
+    ;;
+esac
```
gforsyth, line 20:
```suggestion
case "${pkg_type}" in
  cpp|python)
    ;;
  *)
    rapids-echo-stderr 'Pass "cpp" or "python" as an argument.'
    exit 1
    ;;
esac
```

tools/rapids-get-pr-conda-artifact-github
```diff
@@ -0,0 +1,46 @@
...
+    export RAPIDS_REF_NAME="pull-request/${pr}"
+
+    # Generate the artifact name
+    artifact_name=$(rapids-package-name "conda_${package_type}")
+
+    # Get the GitHub run ID for the PR
+    github_run_id=$(rapids-github-run-id)
+
+    rapids-echo-stderr "Downloading and decompressing ${artifact_name} from GitHub Run ID ${github_run_id} into ${unzip_dest}"
+    gh run download "${github_run_id}" --repo "${RAPIDS_REPOSITORY}" --name "${artifact_name}" --dir "${unzip_dest}"
```
gforsyth, line 43: Isn't this `rapids-download-from-github`?

tools/rapids-get-pr-wheel-artifact-github
```diff
@@ -0,0 +1,53 @@
...
+    export RAPIDS_REF_NAME="pull-request/${pr}"
+
+    # Generate the artifact name
+    artifact_name=$(rapids-package-name "wheel_${package_type}")
+
+    # Get the GitHub run ID for the PR
+    github_run_id=$(rapids-github-run-id)
+
+    rapids-echo-stderr "Downloading and decompressing ${artifact_name} from GitHub Run ID ${github_run_id} into ${unzip_dest}"
+    gh run download "${github_run_id}" --repo "${RAPIDS_REPOSITORY}" --name "${artifact_name}" --dir "${unzip_dest}"
```
gforsyth, line 50: Isn't this `rapids-download-from-github`?

gforsyth (commented):
Thanks for putting this together @VenkateshJaya !

I have some bash simplifications (if such a thing can be said to exist) and a few questions about usage, but I think this tooling will be a big improvement!

tools/rapids-get-pr-conda-artifact-github
ajschmidt8: @gforsyth, so something like this?

```sh
unzip_dest="${RAPIDS_UNZIP_DIR:-$(mktemp -d)}"
```

tools/rapids-get-pr-conda-artifact-github
```diff
@@ -0,0 +1,46 @@
...
+pr="${2}"
+package_type="${3}"
+commit="${4:-}"
+
+# Create a temporary directory for the artifact
+unzip_dest=$(mktemp -d)
+
+# If commit is not provided, get the latest commit on the PR
+if [[ -z "${commit}" ]]; then
+    commit=$(git ls-remote https://github.com/rapidsai/"${repo}".git refs/heads/pull-request/"${pr}" | cut -c1-7)
```
ajschmidt8, line 25: while this command will likely work for public repositories, it wouldn't work for private repositories.

therefore we may want to switch this to use the `gh` CLI instead.

As part of this, we'll need to update the documentation to provide users with instructions on how they can pass in the necessary GitHub secret to fetch workflow artifacts from private repositories.

tools/rapids-get-pr-conda-artifact-github
gforsyth: yeah, exactly

tools/rapids-get-pr-conda-artifact-github
VenkateshJaya: Switched this to use gh cli

tools/rapids-download-from-github
VenkateshJaya: Yes, the expected behavior is that all the required environment variables are set either on the CI image or the `rapids-prompt-local-repo-config` script which is invoked on both the `rapids-download-conda-from-github` and `rapids-download-wheel-from-github` scripts when running locally.
Given that this script is only going to be called by the two parent scripts I specified above and should be used on its own, I think it is a better idea to add a comment in `rapids-github-run-id` where these env-vars are actually being used to help understand how the env-vars are being set. 
Let me know what you think.

tools/rapids-get-pr-conda-artifact-github
VenkateshJaya: The idea here is that downloading PR artifacts will not happen through CI, but instead will be useful for developers reproducing CI locally. In that case, I feel it would be a better idea to confine all functionality to a single script. This also lets us customize PR artifact downloading behavior like choosing a destination path, without changes to `rapids-download-from-github` which will be used by CI.

tools/rapids-get-pr-wheel-artifact-github
VenkateshJaya: The idea here is that downloading PR artifacts is not needed by CI for builds, but instead will be useful for developers reproducing CI locally. In that case, I feel it would be a better idea to confine all functionality to a single script. This also lets us customize PR artifact downloading behavior like choosing a destination path, without changes to `rapids-download-from-github` which will be used by CI.

tools/rapids-get-pr-conda-artifact-github
gforsyth: This script is already calling out to `rapids-package-name` and `rapids-github-run-id`, so developers are going to need to have the `tools` folder on their `$PATH` already.

VenkateshJaya: In that case we would need to modify `rapids-download-from-github` to accept an env-var for the download location.
Maybe @ajschmidt8 can expand more on this, but the ideally we would want to make sure all downloads are done strictly to temporary locations so will have to think about how to incorporate a download location env-var for these scripts

gforsyth: I think if you modify `rapids-download-from-github` to set its `unzip_dest` target as you have in this script (`unzip_dest="${RAPIDS_UNZIP_DIR:-$(mktemp -d)}"`), then you could have all of the specific `rapids-get-...` scripts call `rapids-download-from-github` and let that one script handle either creating the temp dir or using the `RAPIDS_UNZIP_DIR` environment variable.

tools/rapids-download-conda-from-github
```diff
@@ -0,0 +1,23 @@
...
+#      should be downloaded
+set -euo pipefail
+export RAPIDS_SCRIPT_NAME="rapids-download-conda-from-github"
+
+pkg_type="$1"
+# Validate package type argument
+if [[ "${pkg_type}" != "cpp" && "${pkg_type}" != "python" ]]; then
+  rapids-echo-stderr 'Pass "cpp" or "python" as an argument.'
+  exit 1
+fi
```
ajschmidt8, line 15: Can we make this a `case` statement for consistency with some of the existing scripts?

tools/rapids-download-wheels-from-github
```diff
@@ -0,0 +1,14 @@
...
+set -eo pipefail
+
+source rapids-prompt-local-repo-config
+
+pkg_type="$1"
+if [ "${pkg_type}" = "cpp" ] || [ "${pkg_type}" = "python" ]; then
+    # remove pkg_type from args because we handle it in this script
+    shift;
+    pkg_name="$(rapids-package-name "wheel_${pkg_type}")"
+fi
```
ajschmidt8, line 12: I think this logic can be cleaned up a bit more. Now that we've removed the legacy components, can you change this to a `case` statement to validate that the first positional argument is either `cpp` or `python`? Anything else should print an error message and exit with a non-zero code.

tools/rapids-get-pr-conda-artifact-github
ajschmidt8: > The idea here is that downloading PR artifacts is not needed by CI for builds

Using PR artifacts from other repositories in GitHub Actions builds is something that's occasionally done. It lets devs test breaking changes introduced in upstream PRs before the upstream PR is merged.

For instance, the artifacts from an `rmm` PR that has a breaking change can be tested in a `cuDF` PR in order to identify what changes need to be made in `cuDF` to accommodate the `rmm` changes.

Therefore, I think it makes sense to consolidate things as Gil suggested. It's reasonable to assume that anyone calling this script will have the rest of the scripts on their `PATH`.

ajschmidt8 (commented):
Looks good far! Left a few comments, but we're getting closer.

ajschmidt8 (commented):
nice! It's coming along.

I tested some of these scripts locally and found one issue. The `rapids-download-*-from-github` and `rapids-get-pr-*-artifact-github` scripts fails because `rapids-package-name` still returns the `.tar.gz` extension. To fix this, I updated the end of the `rapids-package-name` script to include the following logic:

```diff
diff --git a/tools/rapids-package-name b/tools/rapids-package-name
index eaadc5c..945f2d0 100755
--- a/tools/rapids-package-name
+++ b/tools/rapids-package-name
@@ -76,6 +76,11 @@ if (( append_arch == 1)); then
 fi
 
 # for cpp and python package types, it's a tarball, append .tar.gz and prepend project name
-pkg_name="${repo_name}_${pkg_name}.tar.gz"
+pkg_extension=".tar.gz"
+if [[ "${RAPIDS_NO_PKG_EXTENSION:-}" == "true" ]]; then
+  pkg_extension=""
+fi
+
+pkg_name="${repo_name}_${pkg_name}${pkg_extension}"
 
 echo -n "${pkg_name}"
```

then I was able to set `RAPIDS_NO_PKG_EXTENSION=true` in my local environment and everything started working correctly.

Can you make these changes to your scripts? This change will also require that you update the package extension handling in https://github.com/rapidsai/shared-workflows/pull/297.

ajschmidt8 (approved):

ajschmidt8:
in hindsight, I probably should've suggested using `basename` to make that diff cleaner. but those conditionals and env vars will get removed after this migration is over, so it's a moot point.

tools/rapids-download-from-github
```diff
@@ -0,0 +1,21 @@
...
+pkg_name="$1"
+unzip_dest="${RAPIDS_UNZIP_DIR:-$(mktemp -d)}"
+
+rapids-echo-stderr "Downloading and decompressing ${pkg_name} from Run ID ${github_run_id} into ${unzip_dest}"
```
gforsyth, line 18: Should this be sent to stderr?  Could we use `rapids-logger` instead?

tools/rapids-get-pr-wheel-artifact-github
```diff
@@ -0,0 +1,48 @@
...
+export RAPIDS_SCRIPT_NAME="rapids-get-pr-wheel-artifact-github"
+
+repo="${1}"
+pr="${2}"
+package_type="${3}"
+commit="${4:-}"
+
+# Check if RAPIDS_PY_WHEEL_NAME is set when fetching a wheel
+if [[ -z "${RAPIDS_PY_WHEEL_NAME}" ]]; then
+    echo "The RAPIDS_PY_WHEEL_NAME variable must be set when fetching a wheel artifact"
```
gforsyth, line 26:
```suggestion
    rapids-echo-stderr "The RAPIDS_PY_WHEEL_NAME variable must be set when fetching a wheel artifact"
```

gforsyth (approved):
One suggestion and one question, but nothing blocking.  This looks really nice @VenkateshJaya !

tools/rapids-download-from-github
gforsyth: Ahh, right.  Wow I hate `bash`.  Thanks, AJ!

ajschmidt8: agreed! These scripts started small and simple, but have exploded in complexity.

At some point in might make sense to refactor them into Python. but we can save that conversation for another day.

PR #153, opened 2025-03-24
refactor(conda): remove `rapidsai` channel for non-release builds

gforsyth:
We should have clean separation between `rapidsai` and `rapidsai-nightly`.

Probably should hold off on merging this until the 25.04 release is finished, though, just in case this breaks something.

### DIFF
--- a/tools/rapids-configure-conda-channels
+++ b/tools/rapids-configure-conda-channels
@@ -20,6 +20,11 @@ remove_conda_channel() {
   fi
 }
 
+# Remove release channel if build is not a release build
+if ! rapids-is-release-build; then
+  remove_conda_channel 'rapidsai'
+fi
+
 # Remove nightly channels if build is a release build
 if rapids-is-release-build; then
   remove_conda_channel 'rapidsai-nightly'


AyodeAwe:
> Probably should hold off on merging this until the 25.04 release is finished

@gforsyth we can convert this to draft then?

gforsyth:
> @gforsyth we can convert this to draft then?

Done

gforsyth:
This is blocking strict channel priority for a few projects -- I've worked around it in `kvikio` by removing `rapidsai` in that repo only (in that case, it's because of older `cupy` versions left over in `rapidsai`)

gforsyth:
Fun: https://github.com/rapidsai/gha-tools/pull/107

Looks like this change was made once and then reverted for help with pulling in older versions of dependencies for hotfixes.

That is directly at odds with strict channel priority, so we'll likely need some kind of special-casing for hotfix builds.

AyodeAwe:
@gforsyth we shouldn't have the hotfix problem if we go ahead with this PR.

The reason hotfixes couldn't find older dependencies was because we initially cleaned up older nightlies without respect to versions but now we keep 12 most recent packages of at least 3 most recent rapids versions. See https://github.com/rapidsai/remove-old-packages/pull/12 and https://github.com/rapidsai/remove-old-packages/pull/14.

Therefore I think if we go on and merge this PR to implement separation between `rapidsai` and `rapidsai-nightly` for release and non-release builds, hotfixing should still go on smoothly.

gforsyth:
Awesome -- thanks @AyodeAwe !

PR #157, opened 2025-04-14
refactor(rattler): remove `dask/label/dev` from default rattler channels

gforsyth:
Since `rapids-dask-dependency` now points at stable releases of `dask`, and because including `dask/label/dev` breaks strict channel priority in our CI builds, I'm removing it here.

If any projects DO still require it, we can add it back on a per-project basis, but it is more harm than help as a default here.

### DIFF
--- a/tools/rapids-rattler-channel-string
+++ b/tools/rapids-rattler-channel-string
@@ -1,13 +1,11 @@
 #!/bin/bash
 
 RAPIDS_CHANNEL="rapidsai-nightly"
-DASK_CHANNEL="dask/label/dev"
 NVIDIA_CHANNEL=""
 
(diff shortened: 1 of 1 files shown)


bdice (approved):

jameslamb (approved):
I'm caught up on the discussion and support doing this.

I think this should also be linked to https://github.com/rapidsai/build-planning/issues/166

And that we should follow up with the change @vyasr recommended here: https://github.com/rapidsai/build-planning/issues/166#issuecomment-2749601951

Removing `dask/label/dev` from `dependencies.yaml` and `conda` env YAML files across RAPIDS should help make conda solves (and therefore CI) faster (one less source of repodata to download and consider).

PR #158, opened 2025-04-16
Pipefail telemetry scripts

msarahan:
@bdice and @gforsyth encountered a very obtuse error in building RMM: https://github.com/rapidsai/rmm/actions/runs/14456467636/job/40596452557?pr=1883#annotation:9:2333)

```
rmm install time:
real	0m13.430s
user	0m8.494s
sys	0m1.192s

Error: Dev container exec failed: (exit code: 1)
```

The error was happening here: https://github.com/rapidsai/rmm/actions/runs/14456467636/job/40596452557?pr=1883#step:9:765

```
tee: telemetry-artifacts/build.log: No such file or directory
```

Setting these bash options will make pipelines error out, but we also explicit handle the absence of this folder as a sign that the command output should not be captured to the telemetry folder.

### DIFF
--- a/tools/rapids-telemetry-record
+++ b/tools/rapids-telemetry-record
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -exo pipefail
 
 # Check for at least 2 arguments: filename and command
 if [ "$#" -lt 2 ]; then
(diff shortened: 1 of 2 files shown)


msarahan:
Closing. The fix needed here was in shared-workflows.

PR #160, opened 2025-04-22
Quote command arguments in retry scripts to preserve splitting

### LINKED ISSUE #155: [BUG] `rapids-conda-retry` (and the rest of the *-retry tools) do not preserve argument quoting
`rapids-conda-retry` does not quote the argument string when passing it to conda:
```sh
runConda ${args}
```
(https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-conda-retry#L234)

and 
```bash
timeout --verbose "${timeout_duration}" ${condaCmd} ${args} 2>&1| tee "${outfile}"
```
(https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-conda-retry#L80C5-L80C87)

This causes arguments that contain spaces to not be properly split. This can be reproduced by the following:
```bash
#!/usr/bin/env bash
args=""
# Minimal repro of how rapids-conda-retry recombines the arguments
for arg in "$@"; do
  args="${args} ${arg}"
done

py_program="
import sys

for arg in sys.argv:
    print(arg)
"
python3 -c "${py_program}" ${args}
```
```sh
$ ./bug.bash "{python: [2.7, 3.5], vc: [9, 14]}"
-c
{python:
[2.7,
3.5],
vc:
[9,
14]}
```
This is fixed by:
1. `"${args}"`
2. Or making `args` into an array, and then `${args[@]}"`.

Jacobfaib:
closes #155

### DIFF
--- a/tools/rapids-conda-retry
+++ b/tools/rapids-conda-retry
@@ -77,7 +77,7 @@ condaCmd=${RAPIDS_CONDA_EXE:=conda}
 #    needToRetry: 1 if the command should be retried, 0 if it should not be
 function runConda {
     # shellcheck disable=SC2086
-    timeout --verbose "${timeout_duration}" ${condaCmd} ${args} 2>&1| tee "${outfile}"
+    timeout --verbose "${timeout_duration}" "${condaCmd}" "${args}" 2>&1| tee "${outfile}"
     exitcode=$?
     needToRetry=0
     needToClean=0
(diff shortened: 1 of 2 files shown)


bdice (approved):
@gforsyth Would you take a peek at this too? Should `shellcheck` catch this kind of thing?

gforsyth:
Actually, yeah, I'm going to merge this, then follow up with a separate shellcheck fixes PR.

Thanks for flagging this @Jacobfaib !

Jacobfaib:
Ah shoot, looks like this is breaking downstream https://github.com/nv-legate/legate.internal/actions/runs/14598434789/job/40952515939?pr=2175. I think the better solution is to use array splitting and `$@`:
```bash
args=()

for ...
  args+=("${arg}")
fi

run-foo() {
  foo "$@"
}

run-foo "${args[@]}"
```
I will open a follow-up PR to do this. Sorry for the back-and-forth

PR #163, opened 2025-04-28
Add Github auth script 

VenkateshJaya:
This PR adds a small utility script for Github authentication when a token is not provided, to enable artifact downloads when reproducing CI locally.


tools/rapids-prompt-local-github-auth
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# A utility script that prompts user to authenticate with GitHub in
+# local environments
+
+if [ -z "${GH_TOKEN:-}" ] && [ -z "${GITHUB_TOKEN:-}" ]; then
+  rapids-echo-stderr "No GitHub token detected in environment"
+  rapids-echo-stderr "Please authenticate with GitHub to continue"
+  gh auth login --web --git-protocol https
+fi
```
ajschmidt8, line 9:
```suggestion
fi

```

Can you add a newline to the end of this file to remove the warning indicator from GitHub shown in the screenshot below?

image

I looked up why this is a best practice (I was unaware despite following the convention) and found this SO post which seems to explain it well: https://stackoverflow.com/a/729795

PR #164, opened 2025-04-29
Add fallback to get Github Run ID for local CI reproduction

VenkateshJaya:
As part of our changes to reduce `gh` API calls, we used the env-var `GITHUB_RUN_ID` to identify the correct run to download artifacts from as part of the build process (https://github.com/rapidsai/gha-tools/pull/162)

However, this env-var is only set on Github pipelines, and not on local CI reproduction. This PR adds a fallback to this value, to find the Github run-id for local CI reproduction. 

This PR also exports the GH_TOKEN variable explicitly, to avoid repetitive authentication.


jameslamb (changes requested):
I left some suggestions for your consideration. They're significant enough that I'm choosing to leave a blocking review, `@` me any time for another review once they're addressed.

VenkateshJaya:
@jameslamb Thanks for the suggestions. 

I've written a function in the script, closely mirroring `rapids-retry` and ensuring that `rapids-logger` statements are not written to the output but instead shown on the terminal. 

I have also used elements from the code snippet you linked to refine the Github Auth script, with no exposed tokens and configurable options to use other protocols. However, I am a bit unclear on the scopes required for this, as all we are doing is either view workflow runs or download artifacts. I went ahead with a minimal `repo` and `workflow` scopes, but if you have some ideas in terms of the scopes we'd need for these operations that'd be great!

tools/rapids-prompt-local-github-auth
```diff
@@ -2,8 +2,38 @@
...
+
+    # Handle SSH keygen in container environments to avoid prompts
+    ssh_keygen="$(which ssh-keygen || echo "")"
+    if [ -n "${ssh_keygen}" ] && [ -n "${avoid_gh_cli_ssh_keygen_prompt}" ]; then
+        sudo mv "${ssh_keygen}"{,.bak} || true
+    fi
+
+    # Attempt authentication
+    if ! gh auth login \
+        --web --git-protocol ${git_protocol} \
```
jameslamb, line 22:
```suggestion
        --web --git-protocol https \
```

Since this `git_protocol` variable is only used in one place and its value is hard-coded, let's please just put that hard-coding here to reduce indirection.

tools/rapids-github-run-id
```diff
@@ -5,13 +5,65 @@
...
+    sleep "${sleep_interval}"
+    rapids-logger "rapids-retry: sleep done -> retrying..." >&2
+    
+    output=$("${command}" "$@")
+    retcode=$?
+  done
+  
+  if (( retcode == 0 )); then
+    echo "${output}"
+  fi
```
jameslamb, line 32:
```suggestion
  while (( retcode != 0 )) && (( retries < max_retries )); do
    ((retries++))
    sleep "${sleep_interval}"
    output=$("${command}" "$@")
    retcode=$?
  done
  
  if (( retcode != 0 )); then
    echo "ERROR: '${command}' failed on ${max_retries} attempts. If you think more retries would fix this issue, try setting environment variable RAPIDS_RETRY_MAX to a higher value."
    return ${retcode}
  fi
  return 0
```

I think it'd be confusing to see these log lines claiming this is from `rapids-retry`, when it isn't. For simplicity and stronger guarantees that no log output will get into the output, let's just remove these logging lines and move the diagnostic information to the error case.

jameslamb (changes requested):
Thanks for the changes. Instead of inlining the retrying, I think we should consider adding a flag `--quiet` to `rapids-retry` and using that to silence any logging.

Looking at the GitHub auth changes, I think I might not have been clear... in https://github.com/rapidsai/gha-tools/pull/164#discussion_r2067014677, I was not suggesting literally copying the exact code from the `devcontainers` repo in here. There are things that that code does which might make sense in that context but aren't necessary or desirable in the context of these `gha-tools` scripts. Please see my inline comments with suggestions about simplifying that.

We also should test this, with both public and private repos. I'm especially worried about how things like cloning over HTTPS will work with private repos.

VenkateshJaya:
@jameslamb Thanks for your suggestions. 

I am going to do one thing, which is open a separate PR for the Github Auth script, so that we can discuss and test the HTTPS operations on private repositories. This PR primarily addresses a completely unrelated issue anyways. 

I added the --quiet flag to `rapids-retry` and I am using that to get outputs to `gh run list` in these latest changes.

tools/rapids-retry
```diff
@@ -26,25 +29,68 @@
...
-      rapids-logger "rapids-retry: sleep done -> retrying..."
-
-      ${command} "$@"
-      retcode=$?
+    # Set up file descriptor 3 - point to stderr if quiet, otherwise stdout
+    if [[ "$quiet" == "true" ]]; then
+        exec 3>&2  # FD 3 points to stderr
+    else
+        exec 3>&1  # FD 3 points to stdout
+    fi
```
jameslamb, line 57: In my opinion, adding new file descriptors and doing redirects like this is way too much complexity for this purpose.

In https://github.com/rapidsai/gha-tools/pull/164#pullrequestreview-2807558274, I was envisioning something much simpler, just guarding the 2 `rapids-logger` calls based on whether or not `--quiet` was passed.

```text
if [[ "${quiet}" == "true" ]]; then
    rapids-logger ...
fi
```

bdice: I would agree with this. I don't think FD 3 is needed here.

bdice: Updated in 0685a39.

tools/rapids-retry
```diff
@@ -26,25 +29,63 @@
 #                          seconds, to wait between retries.
 #                          Default is a 10 second sleep
 #
+# Additional options:
+#
+# --quiet                - when specified as the first argument after the command,
+#                          captures the command's output instead of sending it to stdout.
+#                          Logger messages are redirected to stderr.
+#                          Returns the output of the command on success.
+#
```
pentschev, line 38: As per the implementation, `--quiet` needs to be necessarily the first argument if specified. I would either document that or make parsing more robust, I would prefer the latter but I do realize a lot of us are time-constrained so just documenting would suffice, something like this would suffice:

```
$0 [--quiet] <command>
```

tools/rapids-retry
```diff
@@ -26,25 +29,63 @@
...
+    else
+        ${command} "$@"
+        retcode=$?
+    fi
+
+    # Choose rapids-logger FD based on quiet mode: stderr (FD 2) if quiet, stdout (FD 1) otherwise
+    logger_fd="1"
+    if [[ "$quiet" == "true" ]]; then
+        logger_fd="2"
+    fi
```
pentschev, line 67: I think this needs to come before the block above in lines 55-61, since the command runs there first before going to the retry loop to run again.

pentschev (commented):
I've left a few suggestions, I also tested the changes with `pull-request` build type and that worked, so I'm +1 in getting this merged soon.

tools/rapids-retry
bdice: If it passes on the first attempt, there are no calls to `rapids-logger`. The `logger_fd` is only needed in the loop. However, I agree that it makes sense to structure it so that the first attempt and the looped attempts are next to each other. Updated in 857fb9480649c36290f88002734723e2ba6bbcbb.

pentschev (approved):
LGTM, thanks @bdice !

tools/rapids-retry
```diff
@@ -26,25 +29,63 @@
...
+            fd="2"
+        fi
+
+        rapids-logger "rapids-retry: retry ${retries} of ${max_retries} | exit code: (${retcode}) -> sleeping for ${sleep_interval} seconds..." >&"${fd}"
+        sleep "${sleep_interval}"
+        rapids-logger "rapids-retry: sleep done -> retrying..." >&"${fd}"
+
+        # Execute command based on quiet mode
+        if [[ "$quiet" == "true" ]]; then
+            output=$("${command}" "$@")
```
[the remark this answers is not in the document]
jameslamb: I don't think it should. If it's working for you all so far with just `stdout`, let's keep it that way.

jameslamb (approved):
Thanks for the changes @bdice . If this is working for you and @pentschev and gets you unblocked , let's merge it without waiting for me to more carefully re-review.

I need to spend a bit of time familiarizing myself with all the GitHub Artifacts - related new stuff we have here and in `ci-imgs` and `shared-workflows` before suggesting other changes to these things.

tools/rapids-github-run-id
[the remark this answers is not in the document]
jameslamb: > _it indexes the list of results with [0], which might be a bit fragile_

It is fragile, I agree. I already fixed that in #170

> _It relies heavily on some conventions used by RAPIDS_

`branch` and `nightly` are *already* RAPIDS conventions, not generic GitHub Actions concepts. For as long as they continue to produce workflow runs in the same repo, using the same workflow definition files, for an overlapping set of commits, I cannot think of another way to differentiate between those 2 types of runs here, can you?

PR #165, opened 2025-04-30
Add Github Run ID logic locally to pr scripts

VenkateshJaya:
This PR addresses an unwanted implication of using the `GITHUB_RUN_ID` env-var in determining the Github Actions run ID to download build artifacts. (see https://github.com/rapidsai/gha-tools/pull/162)

This env-var is set while using Github Actions, so the scripts take up the ID of the current Github run, instead of determining the ID of the original run from which PR artifacts are to be downloaded. 

This PR localizes all the logic necessary to determine the Run ID of the actions run triggered on the specific commit of the PR, and ensures that all these operations are done in a subshell to prevent any unwanted env-var changes elsewhere.


tools/rapids-get-pr-conda-artifact-github
```diff
@@ -23,20 +23,23 @@ commit="${4:-}"
...
-    export RAPIDS_SHA="${commit}"
-    export RAPIDS_REF_NAME="pull-request/${pr}"
+    RAPIDS_REPOSITORY="rapidsai/${repo}"
+    RAPIDS_SHA="${commit}"
+    RAPIDS_REF_NAME="pull-request/${pr}"
 
     # Generate the artifact name
-    artifact_name="$(RAPIDS_NO_PKG_EXTENSION=true rapids-package-name "conda_${package_type}")"
+    pkg_name="$(RAPIDS_NO_PKG_EXTENSION=true rapids-package-name "conda_${package_type}")"
+    github_run_id=$(gh run list --repo "${RAPIDS_REPOSITORY}" --branch "${RAPIDS_REF_NAME}" --commit "${RAPIDS_SHA}" --json databaseId --jq '.[0] | .databaseId')
```
bdice, line 38: Does this need to use `rapids-retry` in case the GitHub API fails? Will the `$(...)` parse the output correctly if it is retried?
```suggestion
    github_run_id=$(rapids-retry gh run list --repo "${RAPIDS_REPOSITORY}" --branch "${RAPIDS_REF_NAME}" --commit "${RAPIDS_SHA}" --json databaseId --jq '.[0] | .databaseId')
```

tools/rapids-get-pr-wheel-artifact-github
bdice, line 1: Can this logic be refactored into a common file? See https://github.com/rapidsai/gha-tools/blob/main/tools/_rapids-download-from-s3 for an example.

tools/rapids-get-pr-conda-artifact-github
```diff
@@ -23,20 +23,23 @@ commit="${4:-}"
...
+    unzip_dest="${RAPIDS_UNZIP_DIR:-$(mktemp -d)}"
 
-    unzip_dir=$(rapids-download-from-github "${artifact_name}")
-    echo -n "${unzip_dir}"
+    rapids-echo-stderr "Downloading and decompressing ${pkg_name} from Run ID ${github_run_id} into ${unzip_dest}"
+    rapids-retry gh run download "${github_run_id}" --repo "${RAPIDS_REPOSITORY}" --name "${pkg_name}" --dir "${unzip_dest}"
```
gforsyth, line 42: so it seems like this is a workaround for the current behavior of `RAPIDS_BUILD_TYPE=pull-request` in `rapids-github-run-id` -- should we change that instead?

tools/rapids-get-pr-conda-artifact-github
```diff
@@ -23,20 +23,23 @@ commit="${4:-}"
 
 # If commit is not provided, get the latest commit on the PR
 if [[ -z "${commit}" ]]; then
-    commit=$(rapids-retry gh pr view "${pr}" --repo rapidsai/"${repo}" --json headRefOid --jq '.headRefOid')
+    commit=$(git ls-remote https://github.com/rapidsai/"${repo}".git refs/heads/pull-request/"${pr}" | cut -f1)
```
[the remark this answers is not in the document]
VenkateshJaya: Refer to the comment above about the `rapids-retry` PR. (https://github.com/rapidsai/gha-tools/pull/164)

tools/rapids-get-pr-conda-artifact-github
VenkateshJaya: In `rapids-github-run-id`, the default value of the Run ID is the `GITHUB_RUN_ID` environment variable set by Github. Making this change there wouldn't help in the usecase where PR artifacts are needed as part of a Github Workflow run. 

Making this change here would be the best option for both Actions and for reproducing CI locally.

tools/rapids-get-pr-wheel-artifact-github
VenkateshJaya: Yes, that's a great idea. 

Refactored the logic into a common file called `_rapids-get-pr-artifact-github`, mirroring the PR artifact flow used earlier.

bdice (approved):

bdice:
This seems fine to me - @gforsyth please take a final pass when you can, and let's merge. I'd like to test some CI artifacts downstream as soon as this is fixed.

PR #168, opened 2025-05-02
WIP: update Github Authentication flow

VenkateshJaya:
This PR is part of the changes to move Build Artifact storage from S3 (`downloads.rapids.ai`) to the Github Artifact Store (see https://github.com/rapidsai/build-infra/issues/237). 

Made changes to the Github Authentication script, which now checks for authentication status rather than a token, and attempts to authenticate over HTTPS protocol if prior authentication is not found.

PR #172, opened 2025-05-08
update handling of GitHub auth for local, interactive use

### LINKED ISSUE #171: rapids-prompt-local-github-auth should check `gh auth status` as well as environment variables.
https://github.com/rapidsai/gha-tools/blob/bc5bcc1ca9ca650ec1179452de73586c02d14174/tools/rapids-prompt-local-github-auth#L5

I am doing a local reproduction of CI (for https://github.com/rapidsai/cudf/pull/18586) and I was prompted twice for `Please authenticate with GitHub to continue`. I logged in twice, but the second time it said `! You were already logged in to this account`.

Ideally, a maximum of one `gh auth login` flow would be triggered. By checking `gh auth status` before prompting to login, we could know ahead of time whether a token exists in the local environment even if `GH_TOKEN` / `GITHUB_TOKEN` are not set.

jameslamb:
Closes #171

Contributes to https://github.com/rapidsai/build-infra/issues/237

Replaces #168

Some of the scripts in this project use the GitHub CLI. Those require authentication with GitHub.

In CI jobs, that authentication is provided by setting environment variable `GH_TOKEN`. When using the scripts interactively (e.g., when reproducing CI), it'd be helpful to prompt for an interactive login flow if the local environment isn't already authenticated.

This does that, via that following:

* in `rapids-prompt-local-github-auth`, prompts for interactive authentication if not currently auth'd with GitHub (determined via `gh auth status`)
* invokes `rapids-prompt-local-github-auth` in every script that uses the GitHub CLI and which might be called interactively (so excludes `rapids-upload-to-anaconda-github`, for example)

### How I tested this

Logged out of GitHub locally.

```shell
unset GH_TOKEN
gh auth logout

gh auth status
# You are not logged into any GitHub hosts. To log in, run: gh auth login
```

Tried to download some `rmm` conda packages.

```shell
RAPIDS_BUILD_WORKFLOW_NAME="build.yaml" \
RAPIDS_BUILD_TYPE="branch" \
RAPIDS_REPOSITORY="rapidsai/rmm" \
RAPIDS_REF_NAME="branch-25.06" \
RAPIDS_SHA="dbd8cc7a5f2230a21551670106d4bcd5e507441e" \
rapids-download-from-github \
    rmm_conda_python_cuda12_py312_aarch64
```

Got an interactive prompt to login:

Screenshot from 2025-05-08 16-15-21

Followed that flow, then was logged in and saw packages download successfully.

Ran that same `rapids-download-from-github` command again, saw the downloads work and NOT re-prompt for auth.

Logged out again, then set the `GH_TOKEN` environment variable to a GitHub "classic" token ([docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-personal-access-token-classic)) that only had the `repo` scope.

```shell
 export GH_TOKEN=<redacted>
gh auth status
```

Confirming that it only had the `repo` scope:

```text
  ✓ Logged in to github.com account jameslamb (GH_TOKEN)
  - Active account: true
  - Git operations protocol: https
  - Token: ghp_************************************
  - Token scopes: 'repo'
  ! Missing required token scopes: 'read:org'
  - To request missing scopes, run: gh auth refresh -h github.com
```

Running the same `rapids-download-from-github` command as above worked, without re-prompting for auth.

Also tested this in an `rmm` PR:

* PR: https://github.com/rapidsai/rmm/pull/1909
* build link: https://github.com/rapidsai/rmm/actions/runs/14916596953/job/41903518506?pr=1909

### DIFF
--- a/tools/_rapids-get-pr-artifact-github
+++ b/tools/_rapids-get-pr-artifact-github
@@ -20,6 +20,8 @@ if [[ "${package_format}" = "wheel" && -z "${RAPIDS_PY_WHEEL_NAME:+placeholder}"
     exit 1
 fi
 
+source rapids-prompt-local-github-auth
+
 # If commit is not provided, get the latest commit on the PR
 if [[ -z "${commit}" ]]; then
     commit=$(rapids-retry --quiet gh pr view "${pr}" --repo rapidsai/"${repo}" --json headRefOid --jq '.headRefOid')
--- a/tools/rapids-download-conda-from-github
+++ b/tools/rapids-download-conda-from-github
@@ -8,7 +8,6 @@ set -euo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-conda-from-github"
 
 source rapids-prompt-local-repo-config
-source rapids-prompt-local-github-auth
 
 # Validate package type argument
 pkg_type="$1"
--- a/tools/rapids-download-from-github
+++ b/tools/rapids-download-from-github
@@ -6,10 +6,12 @@
 set -euo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-from-github"
 
-if [ -z "$1" ]; then
+if [ -z "${1:-}" ]; then
   rapids-echo-stderr "Must specify input arguments: PKG_NAME"
   exit 1
 fi
+
+source rapids-prompt-local-github-auth
 
 github_run_id="$(rapids-github-run-id)"
 pkg_name="$1"
--- a/tools/rapids-download-wheels-from-github
+++ b/tools/rapids-download-wheels-from-github
@@ -8,7 +8,6 @@ set -eo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-wheels-from-github"
 
 source rapids-prompt-local-repo-config
-source rapids-prompt-local-github-auth
 
 # Validate package type argument
 pkg_type="$1"
--- a/tools/rapids-github-run-id
+++ b/tools/rapids-github-run-id
@@ -5,6 +5,8 @@
 set -euo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-github-run-id"
 
+source rapids-prompt-local-github-auth
+
 # While called by CI, all the environment variables are set by the caller. However when run locally, these environment variables are set by rapids-prompt-local-repo-config
 case "${RAPIDS_BUILD_TYPE}" in
   pull-request)
--- a/tools/rapids-prompt-local-github-auth
+++ b/tools/rapids-prompt-local-github-auth
@@ -1,9 +1,30 @@
 #!/bin/bash
-# A utility script that prompts user to authenticate with GitHub in
-# local environments
+#
+# Checks if the current environment is authenticated to communicate with the GitHub API.
+#
+# If not, prompts for an interactive login to generate short-lived credentials.
+#
+# This exists primarily for interactive use cases, like trying to reproduce CI locally.
+#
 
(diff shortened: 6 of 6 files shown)


tools/_rapids-get-pr-artifact-github
```diff
@@ -20,6 +20,8 @@ if [[ "${package_format}" = "wheel" && -z "${RAPIDS_PY_WHEEL_NAME:+placeholder}"
     exit 1
 fi
 
+source rapids-prompt-local-github-auth
```
bdice, line 23: Yes! I came to this conclusion too: https://github.com/rapidsai/gha-tools/issues/171#issuecomment-2864295152

bdice (commented):
I know this isn't ready for review yet - just noting that you already solved one of my questions elsewhere, so no need to respond on that issue.

tools/_rapids-get-pr-artifact-github
jameslamb: Yeah! Everywhere that we invoke `gh` in a tool that might be used interactively, this should be called.

The `if gh auth status` makes it idempotent (ensures you aren't put through the interactive login workflow multiple times).

tools/rapids-download-from-github
```diff
@@ -6,11 +6,13 @@
 set -euo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-from-github"
 
-if [ -z "$1" ]; then
+if [ -z "${1:-}" ]; then
```
jameslamb, line 9: The use of `set -u` a few lines up means that if you fail to pass any arguments to this script, it never gets through this `if` and emits the informative error message. Defaulting back to null if `${1}` is undefined fixes that.

```shell
rapids-download-from-github
# tools/rapids-download-from-github: line 9: $1: unbound variable
```

tools/rapids-download-wheels-from-github
```diff
@@ -8,7 +8,6 @@ set -eo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-wheels-from-github"
 
 source rapids-prompt-local-repo-config
-source rapids-prompt-local-github-auth
```
jameslamb, line 11: Removing this because this script doesn't directly call the `gh` CLI. These `rapids-prompt-local-github-auth` calls should only be placed in scripts that directly need GitHub auth, to minimize unnecessary calls.

tools/rapids-download-conda-from-github
```diff
@@ -8,7 +8,6 @@ set -euo pipefail
 export RAPIDS_SCRIPT_NAME="rapids-download-conda-from-github"
 
 source rapids-prompt-local-repo-config
-source rapids-prompt-local-github-auth
```
jameslamb, line 11: Removing this because this script doesn't directly call the `gh` CLI. These `rapids-prompt-local-github-auth` calls should only be placed in scripts that directly need GitHub auth, to minimize unnecessary calls.

jameslamb:
Ok this one's ready for review whenever you have time @bdice

tools/rapids-download-from-github
bdice: I observed this too and didn’t realize it was a bug. 🥲

bdice (approved):

tools/rapids-download-from-github
jameslamb: Ha yup! Same thing as https://github.com/conda-forge/arrow-cpp-feedstock/pull/1696, it's easy to trip over stuff like this with `set -u`.

PR #173, opened 2025-05-09
reduce duplication in rapids-get-pr-* scripts

jameslamb:
Contributes to https://github.com/rapidsai/build-infra/issues/237

For a long time now, we've had tools in this project to make it easier to use one project's CI artifacts in another project's CI:

* `rapids-get-pr-conda-artifact`
* `rapids-get-pr-wheel-artifact`

Today, those download files from the `downloads.rapids.ai` S3 bucket. So when the GitHub Actions artifact storage stuff was first rolling out, #152 introduced GitHub-y versions of those:

* `rapids-get-pr-conda-artifact-github`
* `rapids-get-pr-wheel-artifact-github`

Now that all RAPIDS projects are uploading conda packages and wheels to the GitHub Actions artifact store, it's safe to remove those `*-github` scripts and just change the behavior of the others to use the GitHub Actions artifact store instead of S3.

### Is it safe to merge this?

Yes. These scripts are not used in any long-lived CI.

GitHub search: https://github.com/search?q=%2Frapids%5C-get%5C-pr%2F%20AND%20NOT%20is%3Aarchived&type=code

They're only used temporarily on PRs during testing. See the docs where that's described: https://github.com/rapidsai/docs/blob/00878de8b9afead0f8d44221c228fda2dbbbd97e/resources/github-actions.md?plain=1#L186-L188

### How I tested this

Created an `rmm` PR using this branch's `gha-tools`: https://github.com/rapidsai/rmm/pull/1909

And then a `cudf` PR using this branch's `gha-tools` and the packages produced by CI from that `rmm` PR: https://github.com/rapidsai/cudf/pull/18747

Checked logs to confirm that the downloaded `rmm` packages were used:

* `branch-25.06` testing: https://github.com/rapidsai/cudf/pull/18747#issuecomment-2878495784
* `branch-25.08` testing: https://github.com/rapidsai/cudf/pull/18747#issuecomment-2913275669


bdice (approved):
Nice! All this makes sense. Thanks especially for your writeup on "Is this safe to merge?" because I had to think about that for a bit.

PR #177, opened 2025-05-13
use pre-commit in CI, other CI changes

jameslamb:
I've been working in this repo a lot recently, and found myself really wanting the development workflow from so many other RAPIDS repos, where you can run most linting locally with a single command (`pre-commit run --all-files`).

This proposes the following:

* using `pre-commit` to run `shellcheck`, instead of the third-party action https://github.com/ludeeus/action-shellcheck
* adding more linters:
  - `ruff` - linting and auto-formatting Python code
  - `pre-commit/pre-commit-hooks` - for generic stuff like trailing whitespace and ensuring every file ends in a newline
* using the system `python` on GitHub's `ubuntu-latest` instead of invoking `actions/setup-python` to run tests

### DIFF
--- a/.github/workflows/prs.yaml
+++ b/.github/workflows/prs.yaml
@@ -17,11 +17,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Run ShellCheck
-        uses: ludeeus/action-shellcheck@master
-        with:
-          scandir: "./tools"
-          ignore_paths: "**.py"
+      - name: Run pre-commit
+        uses: pre-commit/action@v3.0.1
       - name: Check Permissions
         run: |
           # Executable bit indices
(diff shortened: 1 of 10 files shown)


.github/workflows/prs.yaml
```diff
@@ -37,14 +34,11 @@ jobs:
...
                 echo "${FILE} does not have the correct permissions."
-                echo "Ensure are tools are executable by everyone and try again."
+                echo "Ensure all tools are executable by everyone and try again."
                 exit 1
               fi
             done
           done
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
```
jameslamb, line 47: We only need Python here to run the test code, which is just `pytest` + the standard library. Whatever version happens to be pre-installed on the `ubuntu-latest` runner from GitHub ([3.12.3, as of this writing](https://github.com/actions/runner-images/blob/25b33392eaa1122a494597b09ecfddd8c3cd1fa1/images/ubuntu/Ubuntu2404-Readme.md#L26)) should be fine.

KyleFromNVIDIA (approved):

ajschmidt8 (approved):

PR #184, opened 2025-05-20
Ensure all extra `rapids-download-from-github` output goes to `stderr`

ajschmidt8:
This PR addresses an issue with `rapids-download-from-github`. That script should only output a folder name to `stdout`. All other output should go to `stderr`. This PR addresses that by redirecting the existing output to `stderr`.


jameslamb (approved):
Yep this looks good to me, thank you!

PR #257, opened 2026-05-06
fix(ci): resolve all zizmor findings and add zizmor pre-commit checks

gforsyth:
Similar to upstream changes in `shared-workflows`, this PR cleans up and annotates all of the workflows and adds the `zizmor` linter to make sure changes are checked.

Part of https://github.com/rapidsai/build-planning/issues/275

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,13 +29,16 @@ jobs:
       - name: Compute Next Tag Name
         run: |
           LAST_TAG=$(git describe --tags --abbrev=0 || echo "v0.0.0")
-          NEXT_PATCH_VERSION=$(echo ${LAST_TAG} | awk -F. -v OFS=. '{$NF += 1 ; print}')
-          echo "NEXT_PATCH_VERSION=$NEXT_PATCH_VERSION" >> $GITHUB_ENV
+          NEXT_PATCH_VERSION=$(echo "${LAST_TAG}" | awk -F. -v OFS=. '{$NF += 1 ; print}')
+          echo "NEXT_PATCH_VERSION=${NEXT_PATCH_VERSION}" >> "${GITHUB_ENV}"
       - name: Release
-        uses: softprops/action-gh-release@v1
-        with:
-          files: ${{ env.RELEASE_FILE_NAME }}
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -1,40 +1,44 @@
...
-        with:
-          files: ${{ env.RELEASE_FILE_NAME }}
-          tag_name: ${{ env.NEXT_PATCH_VERSION }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release create "${NEXT_PATCH_VERSION}" "${RELEASE_FILE_NAME}" \
+            --target "${GITHUB_SHA}" \
+            --title "${NEXT_PATCH_VERSION}" \
+            --notes ""
```
jameslamb, line 41: Nice! Yeah if this `gh` CLI call does the same thing as that third-party action, it's definitely preferable.

.github/zizmor.yml
```diff
@@ -0,0 +1,10 @@
+
+rules:
```
jakirkham:
```suggestion
rules:
```

jakirkham (approved):
Thanks Gil! 🙏

Had one nit below. Otherwise LGTM

.github/workflows/release.yaml
gforsyth, line 41: This gets rid of a third-party action in favor of `gh` functionality -- it will require a merge (of this PR) to test that the release happens as expected.

jakirkham: Thanks Gil! 🙏

Agree this is a nice improvement

Do we need any other permissions for this to work?

jameslamb (approved):
Very happy to see one less third-party action dependency!

.github/workflows/release.yaml
gforsyth: I think `github.token` is sufficient?  But we'll find out!

gforsyth:
Ok, going to merge this, then make sure that a new release gets cut

gforsyth:
BOOM: https://github.com/rapidsai/gha-tools/releases/tag/v0.0.201

(that's the good kind of `BOOM`)

PR #273, opened 2026-08-05
Add optional output type variable to `rapids-generate-pip-constraints`

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/build-planning/issues/311


jameslamb (approved):
IMO it'd be better to have another optional positional argument than a new special-purpose shell variable. But fine with either way you decide to go.

